PR #54, opened 2015-01-08
Feature/less restrictive deps

D1plo1d:
Making the deps less restrictive so that exq is compatible with phoenix 0.7.2.


D1plo1d:
w00t! :)

PR #57, opened 2015-04-01
Update uuid dependency to 1.0.0

Nielsomat:
and grab it from hex instead of github.

(Exq is not affected by any of the breaking changes: https://github.com/zyro/elixir-uuid/blob/master/CHANGELOG.md#v100-09-mar-2015-elixir--10.)


j-mcnally:
:+1:

PR #60, opened 2015-05-18
Support an empty namespace

bbhoss:


j-mcnally:
Do we need to handle nil also or is that not a possible value?

akira:
I'd add in a nil in there as well. Looks good besides that.

bbhoss:
I'm not sure what happens if you explicitly add nil, but if you leave the key out it looks like it uses the exq namespace by default

akira:
Merged.

akira:
Thanks!   

Note:  Also updated master to handle nil as well (namespace: nil).

j-mcnally:
Just to document that,

https://github.com/akira/exq/commit/96355b5b1d41643cb02393a0b598bbeb336d5a4e

PR #66, opened 2015-06-22
Allow throttling for number of concurrent workers

akira:
Keep a count on the manager level and throttle if going over this count.
This is a simple solution for now - could be changed once we go to Queue level
concurrency / worker supervision

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,12 +30,14 @@ Then run ```mix deps.get```.
 
 By default, Exq will use configuration from your config.exs file.  You can use this
 to configure your Redis host, port, as well as namespace (which helps isolate the data in Redis).
+The "concurrency" setting will let you configure the amount of concurrent workers that will be allowed, or :infinite to disable any throttling.
 
 ```elixir
 config :exq,
   host: '127.0.0.1',
   port: 6379,
   namespace: "exq",
+  concurrency: :infinite,
   queues: ["default"]
 ```
 
--- a/lib/exq/manager.ex
+++ b/lib/exq/manager.ex
@@ -6,14 +6,23 @@ defmodule Exq.Manager do
   @default_name :exq
 
   defmodule State do
-    defstruct pid: nil, host: nil, redis: nil, namespace: nil, busy_workers: 0,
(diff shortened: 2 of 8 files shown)


akira:
@j-mcnally added some simpler throttling to use until we figure out queue level or supervised throttling

j-mcnally:
gotcha, i dont know if you want to take another look at https://github.com/structtv/exq/tree/concurrency also.

akira:
I was thinking if we went with your other suggestion to use a supervisor with dynamic supervision per Queue and throttle concurrency there - we can probably get by without using ETS and just do it at the supervisor level?  What I wasn't crazy about in the implementation above is that the RedisQueue needs to know about workers. 

Anyways, this was just a temporary step, and was looking at adding it in since it was so simple, until we figure out the above.

Oh, one other thing - for the  "per queue", where should the concurrency configuration go?

akira:
@j-mcnally I was also wondering - in your implementation, if 1 queue is overloaded but the others are free, but busy one gets picked in the shuffle, it will not dequeue anything right? It should probably limit the shuffle to non-busy queues.

If we wanted to go with this implementation (vs queue supervisor), was wondering if we can create a 4-arity method on Queue dequeue_random that has accepts a boolean function that can reject certain queues based on worker status, and maybe pre-filter this before the shuffle.  

Let me know your thoughts..

j-mcnally:
Yeah, im not saying mine is right approach, just wanted to throw out what i had done to address the problem.

I believe if the max concurrency is hit when dequeue_random hits, and a queue has x number of works assigned to it, 

https://github.com/structtv/exq/blob/concurrency/lib/exq/redis_queue.ex#L83 makes a call to can_run which checks how many workers are running per queue.

In the case the dequeue returns nothing or nil, https://github.com/structtv/exq/blob/concurrency/lib/exq/redis_queue.ex#L85

It simple re-runs dequeue again to try to pop something it can run.

akira:
@j-mcnally I took some ideas from your branch but reshuffled it around so we wouldn't need to JSON parse the job each time.  Tried a few things, but ended up doing it from the manager and pass around the queue. Let me know if you had thoughts.

PR #68, opened 2015-07-04
Handle JSON parsing errors in Exq.Stats.remove_process

disbelief:
I was running into an issue that was causing exq to crash when `Exq.Stats.remove_process` was called. The reason for the crash was due to JSON parsing errors when reading from the `processes` set in Redis.

While vanilla Sidekiq only places valid JSON into `processes`, there are a few common Ruby Sidekiq plugins which insert other things. In my case, the [sidekiq-limit_fetch](https://github.com/brainopia/sidekiq-limit_fetch) plugin was inserting string hashes into the `processes` set.

So I've wrapped Poison's non-bang `encode()` and `decode()` functions in `Exq.Json`, and renamed the previous `encode()` and `decode()` functions there to match the underlying Poison function (ie. `encode!()` and `decode!()`)

If this change breaks things for people *(perhaps they were relying on the existing `Exq.Json` API in their own projects?)*, an alternative would be to create a function `Exq.Json.safe_decode()` that calls the non-bang version of `Poison.decode()`, and then only the `Exq.Stats.remove_process` function would need to use it. Let me know if this is preferred to renaming the functions.

Finally, as per the instructions at the bottom of README.md, I attempted to run the tests after my changes and quite a few fail. However after reverting to master and running the tests again, I see the same errors. Therefore I'm pretty sure the tests were already failing and my changes have not introduced anything new.

### DIFF
--- a/lib/exq/job.ex
+++ b/lib/exq/job.ex
@@ -1,10 +1,18 @@
 
 defmodule Exq.Json do
   def decode(json) do
-    Poison.decode!(json)
+    Poison.decode(json)
   end
 
   def encode(e) do
+    Poison.encode(e)
+  end
+
+  def decode!(json) do
+    Poison.decode!(json)
+  end
+
+  def encode!(e) do
     Poison.encode!(e)
   end
 end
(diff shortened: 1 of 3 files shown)


akira:
:+1:  Looks good, thanks!

PR #74, opened 2015-07-18
Lint + harmonize whitespace

optikfluffel:
removed a bunch of [AtomLinter/linter-elixirc](https://github.com/AtomLinter/linter-elixirc) "errors" and some trailing spaces


optikfluffel:
Still the same timing error? Worked again at [my build](https://travis-ci.org/optikfluffel/exq/builds/71530623).

akira:
Yeah, looks like it - I think the container based execution is definitely a bit slower and throwing the timing off.

lib/exq/redis.ex
```diff
@@ -48,10 +48,12 @@ defmodule Exq.Redis do
 
   def keys!(redis, search \\ "*") do
     {:ok, keys} = :eredis.q(redis, ["KEYS", search])
+    keys
```
akira: Stats module is using this, so that would need to be changed as well:

https://github.com/optikfluffel/exq/blob/lint/lib/exq/stats.ex#L200

We don't have an tests for that, of course would be great to have as well.

optikfluffel:
Fixed the `keys!` usage in `stats.ex`, will add some tests when I have a little more time, maybe in the evening or tomorrow.
Would be cool if you'd merge #75 until then, to see if the coverage actually is going up afterwards :smile_cat:

akira:
:+1:

PR #76, opened 2015-07-21
Add scheduled jobs

nicksanders:
Based off #8 I have added a Task Scheduler, 

It uses "scheduled" queues for each "live" queue which are sorted sets, when a scheduled job's time is reached the job is pulled from the "scheduled" queue and put on the "live" queue and then the manager deals with it as normal.

I'm new to Elixir so I thought it best to get some feedback before I went any further.

It is working and api wise I think enqueue_in works but I'm not sure about the format for the time in enqueue_at, currently its a unix timestamp.

At the moment the jobs are scheduled with a 1 second resolution which works for me but I'm not sure how accurate others need the scheduling to be.

I thought my next steps should be to add a setting to enable the scheduler which would default to false and also add tests and documentation for the new functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,10 @@ config :exq,
   password: "optional_redis_auth",
   namespace: "exq",
   concurrency: :infinite,
-  queues: ["default"]
+  queues: ["default"],
+  poll_timeout: 50,
+  scheduler_enable: false,
+  scheduler_poll_timeout: 200,
 ```
 
 ### Concurrency:
@@ -134,6 +137,18 @@ You can also enqueue jobs without starting workers:
 
 {:ok, ack} = Exq.Enqueuer.enqueue(:exq_enqueuer, "default", "MyWorker", [])
 
+```
+You can also schedule jobs to start at a future time:
+You need to make scheduler_enable is set to true
+
+Schedule a job to start in 5 mins
+```elixir
+{:ok, ack} = Exq.enqueue_in(:exq, "default", 300, "MyWorker", ["arg1", "arg2"])
+```
+Schedule a job to start at 8am 2015-12-25 UTC
+```elixir
+time = Timex.Date.from({{2015, 12, 25}, {8, 0, 0}}) |> Timex.Date.to_timestamp
+{:ok, ack} = Exq.enqueue_at(:exq, "default", time, "MyWorker", ["arg1", "arg2"])
 ```
 
 ### Creating Workers:
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ You can also enqueue jobs without starting workers:
 
 ```
 You can also schedule jobs to start at a future time:
-You need to make scheduler_enable is set to true
+You need to make sure scheduler_enable is set to true
 
 Schedule a job to start in 5 mins
 ```elixir
--- a/lib/exq/redis_queue.ex
+++ b/lib/exq/redis_queue.ex
@@ -47,16 +47,12 @@ defmodule Exq.RedisQueue do
       ["RPUSH", queue_key(namespace, queue), job_json]])
   end
 
-  def enqueue_at(redis, namespace, queue, time, worker, args) do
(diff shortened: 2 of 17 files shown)


j-mcnally:
Nick this looks awesome, but tests are failing. I will review with @akira but in the meantime can you check the tests?

j-mcnally:
@nicksanders  this looks awesome, but tests are failing. I will review with @akira but in the meantime can you check the tests?

nicksanders:
Thanks, the tests are passing on OTP 18 but failing on OTP 17.5? I'll look into it tomorrow

akira:
@j-mcnally I reran the build, looks like test passed.  I think it's the containerized Travis that seems to be giving timing related failures.

j-mcnally:
Awsome. checking out the code.

j-mcnally:
this looks like a legit implementation, can i ask what it does to our Sidekiq/Redis compatibility? The fact that its all stored in Redis is a +1

akira:
@nicksanders Looks great, thanks!  

If you did have time, would be great to add some tests, for example testing the scheduled_dequeue stuff, or integration tests..

nicksanders:
@j-mcnally not sure which to be compatible with sidekiq, resque-scheduler, resque-delayed ;)

Looking at Sidekiq's code, they use a single schedule queue so we would have to decode jobs if there are multiple queues but there would be less polling. I'm open to that. It also looks like they use a string for the score not an integer which I can look into.

@akira I have time, its all new to me so might take a couple of evenings ;)

nicksanders:
I have changed it to use a single schedule queue like Sidekiq and also the same scoring so should be compatible.

I'll add some more tests and look at the stats and web ui next.

lib/exq/redis_queue.ex
```diff
@@ -23,10 +32,35 @@ defmodule Exq.RedisQueue do
...
+    enqueued_at = DateFormat.format!(Date.from(time, :timestamp) |> Date.local, "{ISO}")
+    {jid, job_json} = to_job_json(queue, worker, args, enqueued_at)
+    score = time_to_score(time)
+    Exq.Redis.zadd!(redis, scheduled_queue_key(namespace), score, job_json)
+    jid
+  end
+
+  def enqueue_in(redis, namespace, queue, offset, worker, args) do
+    time = Time.add(Time.now, Time.from(offset, :secs))
+    enqueued_at = DateFormat.format!(Date.from(time, :timestamp) |> Date.local, "{ISO}")
```
akira: Just noticed this - can probably just call enqueue_at with the specific time instead of redefining it here

README.md
```diff
@@ -135,6 +138,18 @@ You can also enqueue jobs without starting workers:
 {:ok, ack} = Exq.Enqueuer.enqueue(:exq_enqueuer, "default", "MyWorker", [])
 
 ```
+You can also schedule jobs to start at a future time:
+You need to make scheduler_enable is set to true
+
+Schedule a job to start in 5 mins
+```elixir
+{:ok, ack} = Exq.enqueue_in(:exq, "default", 300, "MyWorker", ["arg1", "arg2"])
```
akira: I'm not sure this is working correctly - running this line seems to execute the job right away.  

I tried enqueueing jobs from Sidekiq scheduler, and that seems to work correctly. Maybe the time offset calculation is not working correctly?

nicksanders: I'm confused ;) it seems to be a bug in Timex but

enqueue_in works for me in my app with Timex 0.13.5
enqueue_in doesn't work in tests with Timex 0.13.5
enqueue_in does work in tests with Timex 0.16.2 but the performance test now takes 50 secs instead of 10 secs

akira:
Great, that works!    :clap: 

Ready to merge, or more WIP?

nicksanders:
Thanks, its ready to merge if you are happy :smile:

I also have a patch waiting to be merged in Timex that fixes the performance issue so you should will be able to upgrade to the latest Timex in the future.

akira:
:+1: merging, thanks!   

Nice find with the Timex issue!  Let me know or issue a PR for when that update is available.

PR #90, opened 2015-09-15
Add support for string host and password. Add test for redis config. …

tuvistavie:
Hi,

I made some changes in redis connection configuration.

* Transform `"hostname"` to `'hostname'` before passing it to eredis when needed.
* Transform `"password"` to `'password'` before passing it to eredis when needed. I think that `password: "optional_redis_auth"`, as written in the README, will fail otherwise.
* Use `|| ''` instead of default argument for the password. With `Config.get(:password, '')`, if `password` is explicitly set to `nil`, for example with someting like `password: System.get_env("REDIS_PASSWORD")`, the default value is not used and the password passed to eredis becomes nil, resulting in a connection error.

I also added some tests for the changes.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -4,7 +4,7 @@ config :logger, :console,
   format: "\n$date $time [$level]: $message \n"
 
 config :exq,
-  host: '127.0.0.1',
+  host: "127.0.0.1",
   port: 6379,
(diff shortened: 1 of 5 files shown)


akira:
:clap: looks great, merging.  Thanks!

PR #96, opened 2015-10-12
Added support for dynamic subscribing and unsubscribing

dl103:
Hey, I implemented the ability for Exq to dynamically subscribe and unsubscribe new queues and a couple test cases along with it. This is my first Elixir commit so let me know if anything needs to be fixed!

### DIFF
--- a/lib/exq/manager/server.ex
+++ b/lib/exq/manager/server.ex
@@ -109,6 +109,21 @@ defmodule Exq.Manager.Server do
     Enqueuer.enqueue_in(state.enqueuer, from, queue, offset, worker, args)
     {:noreply, state, 10}
   end
+  
+  def handle_call({:subscribe, queue}, from, state) do
+    updated_state = add_queue(state, queue)
+    {:reply, :ok, updated_state}
+  end
+  
+  def handle_call({:subscribe, queue, concurrency}, from, state) do
+    updated_state = add_queue(state, queue, concurrency)
+    {:reply, :ok, updated_state}
+  end
+  
+  def handle_call({:unsubscribe, queue}, from, state) do
+    updated_state = remove_queue(state, queue)
+    {:reply, :ok, updated_state}
+  end
 
   def handle_call({:stop}, _from, state) do
     { :stop, :normal, :ok, state }
@@ -180,6 +195,19 @@ defmodule Exq.Manager.Server do
       :ets.insert(work_table, queue_concurrency)
     end)
     {queues, work_table}
+  end
+  
+  defp add_queue(state, queue, concurrency \\ Config.get(:concurrency, 10_000)) do
+    queue_concurrency = {queue, concurrency, 0}
+    :ets.insert(state.work_table, queue_concurrency)
+    updated_queues = [queue | state.queues]
+    %{state | queues: updated_queues}
+  end
+  
+  defp remove_queue(state, queue) do
+    :ets.delete(state.work_table, queue)
+    updated_queues = List.delete(state.queues, queue)
+    %{state | queues: updated_queues}
   end
 
   def default_name, do: @default_name
--- a/lib/exq/manager/server.ex
+++ b/lib/exq/manager/server.ex
@@ -112,17 +112,17 @@ defmodule Exq.Manager.Server do
   
   def handle_call({:subscribe, queue}, from, state) do
     updated_state = add_queue(state, queue)
-    {:reply, :ok, updated_state}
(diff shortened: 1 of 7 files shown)


dl103:
Hmm, let me investigate the test cases later tonight or tomorrow.

akira:
Great, thanks!  Will take a look.   

As far as tests, I think it is the new tests that seem to be failing. Perhaps the issue is with expected return from register / unregister?

test/exq_test.exs
```diff
@@ -117,6 +117,30 @@ defmodule ExqTest do
...
+    wait_long
+    assert_received {:worked, 1}
+    assert_received {:worked, 2}
+    stop_process(sup)
+  end
+  
+  test "unregister queue and run job" do
+    Process.register(self, :exqtest)
+    {:ok, sup} = Exq.start_link([name: :exq_t, port: 6555, namespace: "test", queues: ["q1","to_remove"]])
+    {:ok, _} = Exq.unsubscribe(:exq_t, "to_remove")
```
akira: I think this works if:
```elixir
:ok = Exq.unsubscribe(:exq_t, "to_remove")
```
same for the ```Exq.subscribe``` above.

lib/exq/manager/server.ex
```diff
@@ -109,6 +109,21 @@ defmodule Exq.Manager.Server do
     Enqueuer.enqueue_in(state.enqueuer, from, queue, offset, worker, args)
     {:noreply, state, 10}
   end
+  
+  def handle_call({:subscribe, queue}, from, state) do
+    updated_state = add_queue(state, queue)
+    {:reply, :ok, updated_state}
```
akira: I know it's not documented, but these need to return a timeout after which to run again, something like:

```{:reply, :ok, updated_state, 0}```

dl103: Ah, interesting. Is this because we need the handle_info method to be triggered because a timeout of 0?

test/exq_test.exs
dl103: Great, let me try it!

lib/exq/manager/server.ex
akira: It's using ```{reply,Reply,NewState,Timeout}``` response for GenServer handle_call to set the next timeout call (see http://www.erlang.org/doc/man/gen_server.html).

The other option would have been to use a separate timeout call, but at the time this was more flexible.

test/exq_test.exs
dl103: Using the :ok to match makes it pass that assertion, but it seems to like the process mailbox only has {:worked, 2} in there during the "register queue and run job" test. Manual testing shows that both queues work though, so I will need to investigate this too =O

akira: Yeah, I saw that too. I briefly looked into that, but didn't get far enough to figure it out :smile:

dl103: Investigated a bit more last night and saw that running the new tests before the all the other tests in exq_tests.exs results in a success, but putting it after "start using registered name" will cause the fail. The plot thickens!

akira: I'm not seeing that behavior, can you put up the commits you have?

akira: I think I figured this out - looks like an actual issue with the way things were getting dequeued, see #103 . Can you rebase against master and try this again?

dl103: Tried it against master after your latest merge and it works now. Thanks for the help @akira -- that would've been impossible for me to find on my own. Will push latest changes.

dl103:
Looks like the test have finally passed but the coverage has decreased from one of the recent PR's. I think I goofed up a bit by forgetting a step in the rebasing process -- it ended up lumping your recent changes in with this PR.

akira:
Don't worry about the coverage part.  Not sure what happened with the rebase, you can always try to branch off master and cherry pick your changes back / force push to this branch. 

Also, would be good to add this capability to the README.md. I can add it as well though if you can't get to to that part.

akira:
I tried merging locally and the merge isn't clean, so a rebase / cherry pick will be needed.  Let me know if you have any issues with this or needed me to manually merge this.

dl103:
Let me clean up using cherry picking and anything else that's needed to fix up the branch! I will get back to this tomorrow morning. Sorry for tag back and forth -- hectic week haha.

dl103:
Closing this down and re-opening PR with cleaned branch. Refer to #104

PR #104, opened 2015-10-18
Cleaned dynamic subscription

dl103:
Re-branched dynamic subscription implementation to clean up branch history. Implements dynamic subscribe and unsubscribe actions as well as tests for each.


akira:
:+1: looks good!  Great to have this feature, thanks!

Issue #119, opened 2015-11-21
Update web ui dependencies


akira:
We are using an older version of ember, etc.  Upgrade them to a later version to try to stay more current.

akira:
Done via #121

Issue #116, opened 2015-11-14
UI for Scheduled and Retried jobs


akira:
Scheduled jobs are under ```scheduled``` queue, jobs to retry are under ```retry``` queue.  Add a UI tab for viewing and managing these.

akira:
@nickgal saw your branch , do you need any help with the API or anything?

akira:
Provided by #132 & moved to exq_ui

Issue #82, opened 2015-09-05
UI - allow requeue failed jobs


akira:
As an end user, I want to re-enqueue failed jobs in the UI

## Acceptance Criteria

* Add "Requeue" button in failed jobs UI view for each job
* Clicking on "Requeue" button, requeues failed jobs

akira:
Moved to exq_ui
https://github.com/akira/exq_ui/issues/2

PR #121, opened 2015-11-24
Update webui

nickgal:
I wanted to work on #82 and #116 however I ran into issues getting libsass to compile on osx 10.11 and decided to update the webui dependencies.

Edit: I see that this actually addresses #119


akira:
:clap: awesome - I was having issues pulling the npm / bower dependencies as well, so this is great!

Let me know if you have any questions or issues if you try either of the other two you had mentioned. 

Thanks!

Issue #129, opened 2015-12-19
Supervisor.start_link inside GenServer


benwilson512:
Hey!

I talked to James Fish on the core team and confirmed that there are some real issues with starting children in this fashion: https://github.com/akira/exq/blob/master/lib/exq/manager/server.ex#L44

Specifically:

- "There is less guarantee provided by the supervision tree as the process might exit and the supervisor will not have terminated its children."

- "This can lead to problems if the supervisors children are named because the named children might still exist when a restart occurs higher up the tree (above the process calling start_link in its init/1"

- "you lose some advanced OTP features, like code reloading, as process modules are discovered by walking the supervision tree"

Looking at your particular use case there also doesn't seem to be much need to do that anyway. All the names and configuration options could be just as easily set one level higher in the manager supervisor. The only pid that's passed around is the redis connection, and honestly each process would do better with its own connection or a connection pool anyway.

Happy to do a PR if desired.

akira:
Hi Ben, 

Thanks for the tip!  That makes sense, and I agree that it would be much cleaner to set everything at the higher level. As far as the Redis connection, I think using a Redis connection pool would be a nice improvement as well.  A PR for that would be great! :smile:

benwilson512:
Related to this, why is `:ets` used in this module? https://github.com/akira/exq/blob/master/lib/exq/manager/server.ex#L243 Everywhere :ets is used the process state is also available, meaning you could just as easily store the queues in the process state.

akira:
@benwilson512 Originally the other processes were working with the ETS table, but with more recent changes it fell all in the same process. I was still planning on redoing some of the way state is handled, so I didn't change that back. I was also thinking of adding an ets table at a higher level, perhaps created at the supervisor level to store the active queue list.

benwilson512:
I'll also be fixing https://github.com/akira/exq/blob/master/lib/exq/manager/server.ex#L203. The workers will be started via a simple_one_for_one supervisor so that they aren't just floating about unsupervised.

benwilson512:
You can track progress on https://github.com/benwilson512/exq/pull/1. Right now it's just a PR to a clone of the Redix branch. I'll eventually redirect it here.

akira:
@benwilson512 I just noticed your comment regarding working on fixing the worker supervision - I didn't see it starred on the #140 list so didn't realize you were working on that. I just put out a branch #147 which does this, let me know if you would like me to hold off on merging if your branch already implements some of this.

[...]

PR #135, opened 2015-12-20
redis support supervisor ; change eredis to Redix

zhongwencool:
1. Change eredis  to [redix](https://github.com/whatyouhide/redix)
2. add supervisor for redis( register process's names more elixir)
3. update tests
4. fix all compile warnings

exq_tree

#129


zhongwencool:
#136 is better

Issue #128, opened 2015-12-19
Redis Pool


benwilson512:
Hey!

In looking through the code it appears that all of the genservers started by https://github.com/akira/exq/blob/master/lib/exq/manager/server.ex share a single redis connection.

Have you considered setting up :eredis with poolboy so that the different processes aren't bottlenecked by this single connection?

This would also go a long way towards letting you start the supervisors started in Exq.Manager.Server.init cleanly as children of the manager supervisor instead of directly inside the GenServer init.

Thanks!
- Ben

akira:
:+1: for using a Redis connection pool via poolboy.  Would help with #129 as well as you mentioned.

olivermt:
fishcakez_ on freelang (creator of Connection etc) did a lot of benchmarking against localhost and got 30% less latency with poolboy than just a straight up shared connection.

This kind of makes as the server side of redis is single threaded so you dont really gain anything except adding the poolboy overhead into the loop.

I was the one who PR'd the poolboy stuff to the redix readme.md and this discussion didn't come up until right after that, so I don't know if something has changed.

A simple supervisor that brings Redix back up should probably be all you need anyways.
Yay for less dependencies and less setup code.

akira:
@olivermt do you have a link to this discussion?

PR #136, opened 2015-12-20
redis support supervisor ; change eredis to Redix

zhongwencool:
1. Change eredis to redix
2. add supervisor for redis( register process's names more elixir)
3. update tests
4. fix all compile warnings

exq_tree

#129


.gitignore
```diff
@@ -9,3 +9,6 @@ _build/
 .*.swp
 *.swo
 stdout
+*.iml
+.idea
```
asymmetric: I think this kind of stuff should be left out of this project's `.gitignore`, and that it belongs to a user's individual `$HOME/.gitignore` file. Otherwise we'd have to explicitly ignore temporary files from every editor, in every project.

lib/exq/manager/server.ex
```diff
@@ -1,22 +1,17 @@
 defmodule Exq.Manager.Server do
   require Logger
   use GenServer
-  alias Exq.Stats.Server, as: Stats
   alias Exq.Enqueuer
-  alias Exq.Redis.JobQueue
   alias Exq.Support.Config
-
-  @default_name :exq
```
akira: This changes the default semantics of how to use Exq, correct?  I'm assuming we will have to do this:
```elixir
Exq.enqueue(Exq, "default", "BusyWorker", [])
```
instead of 
```elixir
Exq.enqueue(:exq, "default", "BusyWorker", [])
```
I think that looks good, maybe we can update the docs?   

I can also see there being a method that just defaults to the correct name as well, so a user can skip the first arg if not needed.

akira:
See #138, I had to fix some things to handle Redis connect and disconnect, as well as timeouts. Ideally the system will be able to handle both of these and not shut down when that happens. 

I was trying your branch with and without those changes, and it seems like in both cases there's an issue with handling disconnects / reconnects, and it crashes the system. 

These tests try out those scenarios run this test: 
```
mix test --trace --include failure_scenarios:true
```
Note that the flakey connection test is a bit timing dependent so it doesn't always pass even if things are correct, but it will give you the idea of what to handle. 

To test this manually:
1. Start Redis server
2. Start Exq
3. Stop Redis server
4. Exq should stay up and not crash
5. Start Redis server
6. Should be able to enqueue and run jobs still

zhongwencool:
@akira pass the failure conditions manually  :)   please help review it 

todo update the readme

lib/exq/redis/connection.ex
```diff
@@ -6,20 +6,8 @@ defmodule Exq.Redis.Connection do
...
-    host = Keyword.get(opts, :host, Config.get(:host, '127.0.0.1'))
-    port = Keyword.get(opts, :port, Config.get(:port, 6379))
-    database = Keyword.get(opts, :database, Config.get(:database, 0))
-    password = Keyword.get(opts, :password, Config.get(:password) || '')
-    reconnect_on_sleep = Keyword.get(opts, :reconnect_on_sleep, Config.get(:reconnect_on_sleep, 100))
-    timeout = Keyword.get(opts, :redis_timeout, Config.get(:redis_timeout, @default_timeout))
-    if is_binary(host), do: host = String.to_char_list(host)
-    if is_binary(password), do: password = String.to_char_list(password)
-    {host, port, database, password, reconnect_on_sleep, timeout}
+    {redis_opts, connection_opts} = Exq.Redis.Supervosor.info(opts)
```
akira: This is not being used anymore right?  I noticed misspelling on ```Supervosor.info``` but we can maybe remove this if not used anyways.

akira:
@zhongwencool thanks, failure conditions pass!  

I have a fix for some of the tests:
See https://github.com/zhongwencool/exq/pull/1/files

Besides that it looks good, but maybe there's 1 issue. It is a different failure condition but I think it will cause some confusion, as many users do not have Redis running when trying out Exq.

The condition is:

1) Redis is turned off
2) Start Exq
3) Application should exit (if possible with message saying it could not connect to Redis).

Right now with Redix changes it spins with a lot of error messages and it is hard to see what is the problem.  In previous version with eredis it exits application right away if it is started without a Redis instance so it is easier to see the problem.

lib/exq/redis/connection.ex
zhongwencool: .... 
 remove the connection/1  
done

zhongwencool:
@akira  I checkout the redis connection by runing dequeue_and_dispatch(state)  in init/1    one time
maybe the "ping" way it better?

akira:
Yeah, I think PING is better, probably faster and doesn't hold up init method as much.

zhongwencool:
@akira  done

benwilson512:
So, I think switching to redix is fine, but I don't think the supervision changes help either either #129 or #128. I'm in the middle of working on a PR that pretty significantly revamps the supervision tree to match OTP conventions. Poolboy can be used with either Redix or :eredis I'm fine either way, I just need to know what the redis story is here so I'm not going in the wrong direction.

benwilson512:
I think the best way forward here might be if I or somebody integrated poolboy support in this PR, we merged it, and then I submit the PR for the broader supervision changes.

benwilson512:
Eh scratch that. I'm just rebasing my work on this branch. It'll handle the pooling and supervision stuff.

akira:
@benwilson512 Hadn't realized you started your PR already :smile: Sounds good as far as getting the supervision / pooling stuff in your branch if that works. 

Regarding the ETS question, originally the other processes were working with the ETS table, but with more recent changes it fell all in the same process. I was still planning on redoing some of the way state is handled, so I didn't change that back. I was also thinking of adding an ets table at a higher level, perhaps created at the supervisor level to store the active queue list.

akira:
:+1:  Works great!  Thanks for the fixes!

Merging :dart:

Issue #140, opened 2015-12-23
General Concerns


benwilson512:
Over the course of working through the PR to fix #128 and #129 I've run into a variety of different problems that I'd like to discuss. These problems are unfortunate, because Exq serves a very important role in the community. Many people are coming from a Ruby / Rails background and Exq helps make integrating Elixir with existing Ruby systems much easier. Nonetheless, the problems I present here individually and in aggregate are serious enough to warrant great caution on the part of any who might consider using this library.

I bring these problems up not only because I want Exq to succeed, but because I only have the available time to fix a handful of the most serious ones. A lot of valuable work has gone into this library, particularly around managing the redis association in such a way as to be compatible with Sidekiq. I'm hopeful that this post will bring about awareness of the situation so that others can address what I cannot.

NOTE: Two of these I've already created separate issues for, but I include them here for the sake of completeness.

\* indicates that it will be at least partially addressed in https://github.com/benwilson512/exq/pull/1

### Severe Issues

- \*start_link calls inside GenServer init callback
  - https://github.com/akira/exq/blob/master/lib/exq/manager/server.ex#L41
  - Breaks supervision guarantees about children termination
  - Breaks hot code loading
- Job Workers use GenServer.start calls
  - https://github.com/akira/exq/blob/master/lib/exq/manager/server.ex#L203
  - Totally unsupervised
- GenServers include catch all handle_call and handle_cast clauses
  - Only handle_info should have a catch all clause
- \*Only a single redis connection is shared throughout the entire application, sorta.
  - The redis connection is always unsupervised
  - Certain GenServers are configured to start_link a redis connection in their init function if they aren't passed one by their own `start_link`.
- Jobs restarting is handled not by the scheduler, nor by the manager, nor even the enqueuer, but rather the stats server
  - https://github.com/akira/exq/blob/master/lib/exq/stats/server.ex#L66

The ambiguities around supervision in general and job supervision in particular mean that I have no idea how this library behaves when I call ``./bin/myapp stop` via exrm for example. While it shuts down is it possible for a job to still get started? What happens to running jobs? Clarity here is paramount as I need to be sure that deploying new code to production will not orphan jobs.

### General Issues

These are a hodgepodge of anti-patterns, code smells, and barriers to contribution.

- \*No real top level Supervisor
- \*Every GenServer has its own one_for_one supervisor for no discernable reason.
- \*Hardcoded configuration defaults (not in themselves bad) exist in many disparate locations, often redundantly.
[...]

Issue #37, opened 2014-10-29
Graceful shutdown of workers on exq terminate


akira:
Have manager shutdown workers and wait for a timeout when a terminate is sent to it.  This depends on #5 having a worker pool for workers.

akira:
Implemented by #147

Issue #11, opened 2014-10-23
Should we consider using a supervisor to manage workers rather than the manager server?


j-mcnally:
I am not 100% sure what the difference is but i think there is room for improvement on how we manage workers.

We could use the dynamic supervision described at: http://learnyousomeerlang.com/supervisors

@akira  if you have ideas here please share, but i think we could do better in this category.

akira:
Provided by #147

PR #147, opened 2015-12-31
Worker supervision

akira:
* Dynamic simple_one_for_one supervisor
* Use spawn_link on worker GenServer and trap_exit

Addresses worker part of #140 , #37, and #11


j-mcnally:
:+1:

Excited we are finally getting proper supervisors for workers and redis :)

Also worth tracking #11 on this.

Issue #100, opened 2015-10-15
Ensure Redis Bulk API is used whenever possible


akira:
Make sure we are using Redis bulk API whenever possible, for example in saving stats, etc.

PR #153, opened 2016-01-06
Changes:

sineed:
1. Use `Connection#qp` in `JobStats#realtime_stats`;
2. Add test for `realtime_stats`.

### DIFF
--- a/lib/exq/redis/job_stat.ex
+++ b/lib/exq/redis/job_stat.ex
@@ -11,9 +11,8 @@ defmodule Exq.Redis.JobStat do
   alias Exq.Redis.Connection
   alias Exq.Redis.JobQueue
 
-  def record_processed(redis, namespace, _job) do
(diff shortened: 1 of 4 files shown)


sineed:
Related to #100 
@akira I'm new to Elixir so there might be code smells.

I'm not quite confident in implemented test. I assert only size of generated stats, not the time of job processing. I tried to find a dep like Ruby `Timecop` gem but had no luck.
Also there might be a floating test error: enqueue of two jobs which are not split by the `:timer.sleep` may be done in different seconds. 

Please take a look.

akira:
The changes look good. 

As far as the test, we're trying to avoid :timer.sleep for tests if possible.  I think this would be better as a lower level `job_stat_test`.  You can add failures and successes through `JobStat.record_failure` and JobStat.record_processed` and then just issue the realtime API without having to wait.

sineed:
I've made changes following your advice. Need squash?

akira:
:+1: looks good, thanks!

PR #158, opened 2016-01-15
refactor supervisor structure

zhongwencool:
1. https://github.com/akira/exq/issues/140

https://github.com/akira/exq/issues/129
exq

2. don't put @doc on private function.


zhongwencool:
Please help review this  :) @akira

akira:
Thanks!  Will try to take a look, maybe this weekend.   Not sure if @benwilson512 also had any thoughts or existing work. 

### Comments
Just some initial thoughts, looking at the code organization there is a lot of code that can be split out from `lib/exq/manager/supervisor.ex`.  Also, it seems like this is no longer the manager supervisor but a high level supervisor now?

Maybe this is to be done in the future or different PR, but as far as the queue list, at some point maybe we should split out the queue list into a separate agent / process /ets table (which ever makes the most sense) that is supervised by the main supervisor that everyone can access? That way if anything crashes, the queue list isn't affected - since it can be modified by `subscribe` / `unsubscribe`.  Let me know your thoughts.

### Restart functionality
I tried this branch out quickly but it does not look like the restart is working correctly.  For example in iex, doing a bogus command:
```
Exq.Api.remove_failed(Exq.Enqueuer, "blahblahblah")
```
This causes the process to crash and application to shut down.  Also the same if you shut down Redis while the manager is up. 

Running the failure scenarios may be helpful  (I know there are false negatives, but maybe to get idea of what to handle too):
```
mix test --trace --include failure_scenarios:true --no-start
```

zhongwencool:
@akira 
## Why the restart  not working? 
Reason:  Enqueue and Manager supervisor will try to restart Redis.Client at the same time.  

1.1 All gen_server worker under Manager supervisor;

1.2 Exq.Enqueuer crash by 
 ```elixir
  Exq.Api.remove_failed(Exq.Enqueuer, "blahblahblah")
```
 It terminate at https://github.com/akira/exq/blob/master/lib/exq/enqueuer/server.ex#L195
it will run Redix.stop(pid)

1.3 Redix.stop(pid) exit normal will make Exq.Redis.Client  woker stop.

1.4 Manager supervisor find Enqueuer stop ,it will restart it 
Here is the problem: https://github.com/akira/exq/blob/master/lib/exq/enqueuer/server.ex#L47
Enqueue  init will check Redis.Client it alive,  it die by 1.3, so Enqueue will restart Redis.Client.

1.5 Almost at the same time, Manager supervisor find Redis.Client stop ,it will restart it, but it start by 1.4,
it will failed by already_started, so will retry for 20 times and the reach max retry time
Exq will shutdown....

So It there any reason for start Redis.Client in Enqueue init?

Maybe we should not let Enqueue control Redix.Client (don't start and stop it)?

## the queue list problem.

  I just thought ETS in a supervisor,  update_counter will just O(1).

**Comment by zhongwencool** [comment]
Here is the sasl crash log   :) 
[crash_report.txt](https://github.com/akira/exq/files/91905/crash_report.txt)

File: lib/exq/manager/server.ex [hunk 1]
```diff
@@ -367,13 +329,11 @@ defmodule Exq.Manager.Server do
...
-  defp check_redis_connection(redis, opts) do
+  # Check Redis connection using PING and raise exception with
+  # user friendly error message if Redis is down.
+  defp check_redis_connection(opts) do
     try do
-      {:ok, _} = Exq.Redis.Connection.q(redis, ~w(PING))
+      {:ok, _} = Exq.Redis.Connection.q(opts[:redis], ~w(PING))
     catch
       err, reason ->
         opts = Exq.Redis.Supervisor.info(opts)
```
akira: This module is no longer defined

akira:
For Enqueuer starting Redis that is for API - we can remove this after #159.
For now, can we do this?
```elixir
    redis_owner = case Process.whereis(opts[:redis]) do
      nil ->
        {redix_opts, connection_opts} = Exq.Manager.Supervisor.redix_opts(opts)
        Redix.start_link(redix_opts, connection_opts)
        true
      _ ->
        false
    end
    state = %State{redis: opts[:redis],
                   redis_owner: redis_owner,
                   namespace: namespace}
```
This will put whether it manages Redis in the state - as it was not recording it correctly.

akira:
Actually for Enqueuer, I think Exq.Enqueuer.Supervisor should manage Redix same as top level supervisor

zhongwencool:
Exq.Enqueuer.Supervisor should manage Redix same as top level supervisor

Yeah!  just working on that way. but the test failed a lot ,I will fixed and let your know :)

zhongwencool:
add a configuration 
```elixir
name: UserDefineName
```
we can
```elixir
 {:ok, ack} = Exq.enqueue(UserDefineName, "default", "MyWorker", ["arg1", "arg2"])
```
userdefined

akira:
Looks much better!  I like the opts split out, I was thinking of something like that.  Is this ready to merge?

zhongwencool:
One more thing~
Sometime I find it's hard to find Support module.
We name /stats/server       =>   Exq.State.Server
but support/config  => Exq.Support.Config

Maybe rename all Exq.Support.XXX to Exq.XXX.Support   will be better?

More  elixir   
helpers/opts =>       Exq.Opts.Helper     :)

zhongwencool:
Done~  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

akira:
Merging!  :clap: :clap: :heart:

Issue #63, opened 2015-05-24
RFC: Middleware support


j-mcnally:
I think we have a pretty good base for a lot of use cases.

Personally I have branches that add specific functionality that I want but might not be perfect for everyone.

I would like to explore some sort of contrib libraries to add functionality without maintaining additional forks or shoe horning them into main line.

This proposal is to add hooks into exq mainline that would allow for 3rd party libraries to add extensions.

I'd love thoughts related to how we can build a system to allow middleware similar to what sidekiq supports.

akira:
Agreed, how about doing something similar to Plug / Phoenix pipeline - to define a middleware API and way to configure the middleware chains?

akira:
Would also be ok with a simpler rack like solution for first version if other way is too much work.  I think Sidekiq does this in the processor and runs it through the middleware chain before dispatching the final worker.

optikfluffel:
+1 for something Plug-like

akira:
Provided by #160

PR #160, opened 2016-01-17
Middleware support

sineed:
The feature is intended to extend worker functionality. `Exq.MidllewareServer` stores module names which will be used as middlewares. Some existing functionality inside `Exq.Worker.Server` has been extracted into their own middlewares.


sineed:
@akira this PR is not ready, I have some questions/thoughts and need your advice.

1. First of all I think that I need to rebase from https://github.com/akira/exq/pull/158.
2. I have extracted some functionality in `Exq.Worker.Server` into middlewares to show how middleware can be used. Maybe names are not perfect.
3. I thought that each middleware will be implemented as `GenServer`. For now it is just a module and is evaluated inside worker process.
4. Should I implement a test case that illustrates middleware usage?
5. Documentation is a work in progress.

sineed:
Related to #63

akira:
Overall I think this approach can work.  I like splitting different functions into middleware, that is great for cleaning up the code! 

Some comment: 

* I don't think we should pass the state from the worker straight into the middleware. I think we should pass a data structure similar to [Plug](http://www.phoenixframework.org/v0.11.0/docs/understanding-plug) which the middleware can then call module functions on to either get data out of (like job information) or control the chain. 
* We can use the method above to be able to stop the job execution in certain cases (for example delaying execution to later, as some Sidekiq middlewares do).
* We should define a behavior similar to the [Plug Behavior](https://github.com/elixir-lang/plug/blob/master/lib/plug.ex) that any middleware should implement. 
*Middleware as modules (vs GenServer) is fine.
* :+1: for a test case illustrating usage.
* Maybe this item is more for discussion, but how would we configure additional middleware as part of a complete project? Can it be configured through config.exs or somewhere else?

sineed:
> I don't think we should pass the state from the worker straight into the middleware. I think we should pass a data structure similar to Plug which the middleware can then call module functions on to either get data out of (like job information) or control the chain.

I added `Exq.Middlewares.Pipeline` structure to hold data inside middleware.

> We can use the method above to be able to stop the job execution in certain cases (for example delaying execution to later, as some Sidekiq middlewares do).

I've introduced halting of middleware execution. It is similar to `Plug` functionality except I don't log halt. Should I do logging here?

> We should define a behavior similar to the Plug Behavior that any middleware should implement.

Done.

> :+1: for a test case illustrating usage.

Done. I've also added example of LoggerMiddleware similar to [Sidekiq solution](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/middleware/server/logging.rb).

> Maybe this item is more for discussion, but how would we configure additional middleware as part of a complete project? Can it be configured through config.exs or somewhere else?

I'm not sure that this must be in `config/config.exs` file. If so, I see the following implementation:

```elixir
  config :exq,
    middlewares: [
      [:push, MyMiddleware],
      [:insert_after, Exq.Middlewares.Job, MyMiddleware2],
      [:insert_before, Exq.Middlewares.Stats, MyMiddleware3]
    ]
```

I'm not sure that this is a valid configuration. If this is an invalid config then confguring must be done throught explicit API calls:

```elixir
     Exq.Middlewares.Server.push(middlewares_pid, MyMiddleware)
     Exq.Middlewares.Server.insert_after(middlewares_pid, Exq.Middlewares.Job, MyMiddleware2)
     Exq.Middlewares.Server.insert_before(middlewares_pid, Exq.Middlewares.Stats, MyMiddleware3)
```

lib/exq/middlewares/pipeline.ex
```diff
@@ -0,0 +1,27 @@
...
+  def assign(%Pipeline{assigns: assigns} = pipeline, key, value) when is_atom(key) do
+    %{pipeline | assigns: Map.put(assigns, key, value)}
+  end
+
+  def halt(%Pipeline{} = pipeline) do
+    %{pipeline | halted: true}
+  end
+
+  def assign_worker_state(pipeline, worker_state) do
+    pipeline = assign(pipeline,     :redis, worker_state.redis)
```
akira: Can use pipeline |> to clean some of this up :smile:

test/middlewares_test.exs
```diff
@@ -0,0 +1,179 @@
...
+  end
+
+  defmodule MyMiddleware do
+    @behaviour Exq.Middlewares.Behaviour
+
+    import Exq.Middlewares.Pipeline
+
+    def before_work(pipeline) do
+      send :middlewarestest, :before_work
+      pipeline = assign(pipeline, :process_info, 1)
```
akira: don't need pipeline variable definition (also repeated in a few places above/below)

lib/exq.ex
```diff
@@ -24,6 +24,7 @@ defmodule Exq do
       worker(Redix, [redis_opts, connection_opts]),
       worker(Exq.Stats.Server, [opts]),
       worker(Exq.Enqueuer.Server, [opts]),
+      worker(Exq.Middlewares.Server, [opts]),
```
akira: Can we change `Middlewares` everywhere to `Middleware`?  I believe `Middleware` is also the plural.

lib/exq/middlewares/server.ex
```diff
@@ -0,0 +1,128 @@
...
+##===========================================================
+
+  defp middleware_chain([], pipeline) do
+    pipeline
+  end
+  defp middleware_chain(_modules, %Pipeline{halted: true} = pipeline) do
+    pipeline
+  end
+  defp middleware_chain([module|modules] = a, pipeline) do
+    modules |> middleware_chain apply(module, pipeline.event, [pipeline])
```
akira: This is a bit confusing with space before apply

lib/exq.ex
```diff
@@ -24,6 +24,7 @@ defmodule Exq do
...
       supervisor(Exq.Worker.Supervisor, [opts]),
       worker(Exq.Manager.Server, [opts]),
```
akira: @zhongwencool I just noticed this - should the manager be higher up the chain?  Seems like it should be stopped only after the workers done.

test/middlewares_test.exs
```diff
@@ -0,0 +1,179 @@
...
+      send :middlewarestest, :after_processed_work_halted
+      halt(pipeline)
+    end
+
+    def after_failed_work(pipeline) do
+      halt(pipeline)
+    end
+  end
+
+  defmodule LoggerMiddleware do
```
akira: Awesome!  Can we make this a module outside the test directory? I was thinking this would be great to have as actual middleware to use!

lib/exq/middlewares/manager.ex
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def after_failed_work(pipeline) do
+    notify(pipeline)
+    pipeline
+  end
+
+
+  defp notify(pipeline) do
+    case is_alive?(pipeline.assigns.manager) do
```
akira: I think with supervisor order above we don't need the `alive?` checks so can probably remove a bunch of code from here (and methods below).

lib/exq/middlewares/stats.ex
```diff
@@ -0,0 +1,46 @@
...
+  def before_work(pipeline) do
+    {:ok, info} = Stats.add_process(pipeline.assigns.stats, pipeline.assigns.namespace,
+      pipeline.worker_pid, pipeline.assigns.host, pipeline.assigns.job_json)
+
+    pipeline = assign(pipeline, :process_info, info)
+    pipeline
+  end
+
+  def after_processed_work(pipeline) do
+    case is_alive?(pipeline.assigns.stats) do
```
akira: I think with supervisor order above we don't need the `alive?` checks so can probably remove a bunch of code from here (and methods below).

akira:
1. I think this is valid syntax
```elixir
config :exq,
    middlewares: [
      [:push, MyMiddleware],
      [:insert_after, Exq.Middlewares.Job, MyMiddleware2],
      [:insert_before, Exq.Middlewares.Stats, MyMiddleware3]
    ]
``` 
We can always add this support as future issue.
2. I'm curious if running the pipeline as a module from the worker is faster than a GenServer (the GenServer could potentially be bottleneck since it's 1 process running for all the jobs). The performance test has the number for that to run and it's a bit more for that (I think 1.3 sec vs 1.0 sec). We can alslo add this investigation as future issue as well.

sineed:
@akira I've made some changes based on your remarks. 
>I'm curious if running the pipeline as a module from the worker is faster than a GenServer

I thought that it could be better to have a GenServer because it has an ability to dynamically add new middleware. I don't have a real use case, it is just my opinion.

Will be glad to investigate this moment :innocent:

akira:
> @akira I've made some changes based on your remarks. 
Thanks, those look good! :+1: 

> I thought that it could be better to have a GenServer because it has an ability to dynamically add new middleware. I don't have a real use case, it is just my opinion.
> Will be glad to investigate this moment :innocent: 

If you do investigate this, one option I was thinking is still keep the GenServer for storing the middleware, but then fetch it at the worker and then apply it as a module method once it's fetched.  Also note that middleware GenServer should handle "crash" - right now it won't have any middleware if it crashes and gets restarted?  You can try with `:observer.start` and "kill" middleware server.  You may need to setup the middleware default somewhere else (in init method for GenServer or at the supervisor level).

sineed:
Made some changes:
* Properly initialize state of `Exq.MIddleware.Server` after restart
* Send only one message from `Exq.Worker.Server` to `Exq.Middleware.Server`. It is done before `before_work` stack call.

lib/exq/middleware/job.ex
```diff
@@ -0,0 +1,39 @@
...
+    |> assign(:worker_module, String.to_atom("Elixir.#{mod}"))
+  end
+
+  def after_processed_work(pipeline) do
+    remove_job_from_backup(pipeline.assigns)
+    pipeline
+  end
+
+  def after_failed_work(pipeline) do
+    remove_job_from_backup(pipeline.assigns)
```
akira: Can we swap these? 
```elixir
retry_or_fail_job(pipeline.assigns)
remove_job_from_backup(pipeline.assigns)
```

akira:
:clap::clap: Nice work!  After swapping those two above, I'm good with merging.  Ready to merge on your end?

sineed:
I've done swapping plus some code cleanups. Need squash?

akira:
Yeah, can you squash?

Also, I think it's really awesome that we have middleware now! It would be cool to configure it in an easy way if you don't necessarily need them to be dynamic.   What do you think about adding an optional setting in config.exs that will pick up the middleware chain to override defaults?

```elixir
middleware: [Exq.Middleware.Stats, Exq.Middleware.Job, Exq.Middleware.Manager, Exq.Middleware.Logger]
```

sineed:
Yes, it is a good idea, moved default middleware to `config.exs` file.
Also squashed.

PR #161, opened 2016-01-18
Reset config after each test in test/config_test.ex

sineed:


sineed:
I found errors in test suite which occur after  test in config_test. Resettting config in each test (especially in the last test) fixes this problem

zhongwencool:
Woo~   :+1: :+1:

zhongwencool:
@sineed  Thanks

PR #162, opened 2016-01-18
Use Redix.pipeline/3 in JobQueue.scheduler_dequeue/3

sineed:


sineed:
Related to #100

lib/exq/redis/job_queue.ex
```diff
@@ -161,17 +161,23 @@ defmodule Exq.Redis.JobQueue do
...
-        {:ok, jobs}      ->
-          deq_count = scheduler_dequeue_requeue(jobs, redis, namespace, schedule_queue, 0)
-          deq_count + acc
-        {:error, reason} ->
-         Logger.error("Redis error scheduler dequeue #{Kernel.inspect(reason)}}.")
-         acc
-      end
-    end)
+    queues = schedule_queues(namespace)
+    commands = Enum.map(queues, &(["ZRANGEBYSCORE", &1, 0, max_score]))
```
akira: Does this handle empty case when there are no queues?  If someone is using "subscribe" and "unsubscribe" that could be temporarily true.

akira: Actually never mind, for scheduler there will always be queues.

akira:
:+1:

PR #166, opened 2016-01-24
Run through linter and cleanup a few cases

andrewvy:
Used [Credo](https://github.com/rrrene/credo) to do this.

I've found that the code is fairly clean already. :+1: 

One interesting part is in `lib/exq/support/job.ex`, `to_json` creates a HashDict but doesn't use it for converting to JSON.

(sidenote: my first elixir-based PR! :tada:)


akira:
:+1: Awesome, thanks for the cleanup!    Good catch with `to_json` one!

Issue #159, opened 2016-01-16
Move API functions out of Enqueuer 


akira:
Right now the web front end uses the Enqueuer to get stats information.  This should be moved somewhere else.

PR #168, opened 2016-01-25
Move API from Enqueuer

sineed:
Related to #159


akira:
I tried this and had an issue:
```
iex -S mix
> :application.stop(:exq)
> Exq.Enqueuer.Server.start_link(name: ExqE)
> Exq.enqueue(ExqE.Enqueuer, "default", Worker, [])
> [info]: Error enqueueing -  {:noproc, {:gen_server, :call, [Exq.Redis.Client, {:commands, [["SADD", "exq:queues", "default"], ["LPUSH", "exq:queue:default", "{\"queue\":\"default\",\"jid\":\"6032b840-7238-4a0a-abc5-00bb4ada23c5\",\"enqueued_at\":1453873619803,\"class\":\"Worker\",\"args\":[],\"retry\":true}"]]}, 5000]}}
```
* From above seems like Redis is not started  
* I saw the supervisor was taken out of Enqueuer.  I wonder if there is a way to reuse the main app supervisor both for Enqueuer / API by just pass in some args to only start enqueuer, or only start the API, etc?  Let me know what you think. 
* Wonder if there's a better name than `web`?   As it's not really a web server, it's more of an API / Info process.

sineed:
* It is my fault, I've fixed it and added a test that checks an issue.
* I think this is a great idea. I think of `mode` term: there are 3 modes:
  * `default` mode. Is started by `Exq.start_link`. Supervision tree: 
  *  `enqueuer` mode. Is started by `Exq.start_link(mode: :enqueuer)`. Supervision tree: 
  * `api` mode. Is started by `Exq.start_link(mode: :api)`. Supervision tree:

  The bad thing is that you need to specify different names when starting in different modes. And each supervisor starts its own redis process.
  What do you think about my implementation?
* Reverted rename of `Api`

lib/exq.ex
```diff
@@ -15,23 +15,8 @@ defmodule Exq do
...
-      worker(Exq.Manager.Server, [opts]),
-      worker(Exq.Stats.Server, [opts]),
-      worker(Exq.Enqueuer.Server, [opts]),
-      supervisor(Exq.Worker.Supervisor, [opts])
-    ]
-
-    if opts[:scheduler_enable] do
-      children = children ++ [worker(Exq.Scheduler.Server, [opts])]
-    end
+    unless opts[:mode], do: opts = Keyword.put(opts, :mode, Config.get(:mode))
```
akira: I don't think we need this.  Can we do something like this in opts:
```
mode = opts[:mode] || Config.get(:mode)
```
And set the mode as :default on Config.ex?

sineed:
Made a change regarding to setting mode. Its default value is already set

akira:
:+1: Looking good! 

I'm going to add Api process to the default startup mode as well, as it's only one more, and I can see wanting to access those stats.

PR #171, opened 2016-02-02
Don’t add exq supervisor inside phoenix

zhongwencool:
https://hexdocs.pm/exrm/extra-common-issues.html

If you are running into issues with your dependencies missing their
dependencies, it’s likely that the author did not put the dependencies
in the :application block of their mix.exs. You may have to fork, or
issue a pull request in order to resolve this issue. Alternatively, if
you know what the dependency is, you can put it in your own mix.exs,
and the release process will ensure that it is loaded with everything
else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,16 +236,14 @@ You can then create a module that implements the middleware behavior and defines
 
 ## Using with Phoenix and Ecto
 
-If you would like to use Exq alongside Phoenix and Ecto, you will need to add Exq to your supervision hierarchy so that the Ecto Repo is available by the time jobs start processing. To do this, edit your application file and add a supervisor module. For example, if we have an application called ```HelloPhoenix```, you would edit ```lib/hello_phoenix.ex``` and add the Exq supervisor:
+If you would like to use Exq alongside Phoenix and Ecto, for example, if we have an application called ```HelloPhoenix```, you would edit ```lib/hello_phoenix.ex``` :
 ```elixir
     children = [
       # Start the endpoint when the application starts
       supervisor(HelloPhoenix.Endpoint, []),
       # Start the Ecto repository
       worker(HelloPhoenix.Repo, []),
-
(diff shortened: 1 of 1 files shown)


akira:
The reason I suggested adding into the Phoenix supervision is due to Ecto dependency.  Most likely the jobs will be accessing the database, and they would get errors processing jobs until the Ecto.Repo is started.  

```
      worker(HelloPhoenix.Repo, []),
      supervisor(Exq, [])
```
Adding it via the app we would run into this problem right?

zhongwencool:
@akira Yes your are right, this problem may happen  occasionally.... 
Sorry for that, because I using this way in production env,  it just work well.

But how to solve the release problem?    which tool do you make release?

PS: [can't guarantee the application's staring order](https://groups.google.com/forum/#!topic/elixir-lang-talk/je1cYUyDi8g)

akira:
If we create an exq_ecto application - I think it still will not work, because it needs to be started after `worker(HelloPhoenix.Repo, []),`. Exq will be after `ecto` but the `HelloPhoenix.Repo` will still not be started. Right?

One option is to not start Exq on the application startup, and start it manually (like Ecto Repo).  Even can do a similar pattern like this (so you can even have multiple Redis connections if you want):
```
defmodule HelloPhoenix.Exq do
  use Exq.System,config: :hello_phoenix_exq
end
```
Or the other option is to set something in config to not start it (`exq -> start: false`), and call `Exq.start_link(startup: true)`


Will have to think about it more, what do you think?

shamanime:
I also needed to do this change for the app to run in production. Removed the supervisor and added to applications.

akira:
I modified the documentation with #175 to add `:exq` to application list and remove the part about adding it to supervisor list. That could still be an option to use in the future, but it would require some tweaks, so will leave this out until that part is fully flushed out.  See issue #176 for tracking.

PR #180, opened 2016-03-08
Updates for Timex 2.0.0

### LINKED ISSUE #179: new Timex version throws constant warnings
New app with no other dependencies gives a constant `warning: Timex.Time.to_secs/1 is deprecated, use to_seconds/1 instead` message in console.

Looks like Timex updated to 2.0 in the last day or so. Was able to recreate here: https://github.com/colbyh/test_exq

screen shot 2016-03-07 at 1 13 45 pm

colbyh:
- Timex 2.0 doesn't support `Date.universal` and is replaced with `DateTime.universal`
- DateTime formatter was moved and the DateFormat module no longer exists
- Cleans up a bunch of warnings about secs/seconds, usecs/microseconds

fixes #179

### DIFF
--- a/lib/exq/middleware/logger.ex
+++ b/lib/exq/middleware/logger.ex
@@ -24,7 +24,7 @@ defmodule Exq.Middleware.Logger do
 
 
   defp delta(%Pipeline{assigns: assigns}) do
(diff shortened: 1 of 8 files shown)


akira:
:+1: Thanks for the fix!

PR #181, opened 2016-03-12
Allow poison 2.

tuvistavie:
Hi,

I was trying to upgrade my app to Poison 2 and ran into a dependency issue.
The API used in `Exq.Support.Json` should work with both 1.x and 2.x, so I think
it is ok to relax the dependency.
I also updated the lock file to use 2.x in the tests, but let me know if you would rather
stick with 1.x there.

Thanks!


akira:
:+1: thanks!

PR #189, opened 2016-05-10
Fix `EnqueuerApi#enqueue_at` offset time units doc

rranelli:
Change time unit from milliseconds to seconds.

As we can see in `exq/lib/exq/redis/job_queue.ex:59`, the provided
offset is multiplied by `1_000_000` at treated as a value in
`microseconds`.

(I actually found out that by waiting 10 minutes for a job to process in
production :[ )


akira:
👍  Thanks for catching that!

PR #197, opened 2016-07-05
add support for worker specific retry count

ananthakumaran:
sidekiq also allows a integer value instead of boolean for the retry key(https://github.com/mperham/sidekiq/wiki/Advanced-Options#workers). Incase of integer, it specifies the max retry count for the specific worker.


akira:
👍 👍  looks good. 

Would be nice to have unit tests perhaps calling the `retry_or_fail_job` method with the different retry options, if that's something you had time for. 

Thanks!!

PR #198, opened 2016-07-05
Throttler integration

wli0503:
##### Base functions for Throttler middleware

- Add a new flag `terminated` in `Exq.Middleware.Pipeline` to halt the execution of the pipeline AND the worker
- Add `zcount!` and `sismember!` in `Exq.Redis.Connection` as utility function
- Add logic to handle `terminated` flag in pipeline in `Exq.Worker.Server`
- Expose `Exq.Middleware.Job.remove_job_from_backup/1` as public function instead of private
- Add tests for `terminated` flag as well as `zcount!` and `sismember!` in `Exq.Redis.Connection` module

Note: @akira I add a bit logic into Exq.Worker.Server to handle the termination of a job. The expected behaviour of this flag, when set to true, is to halt the execution of the pipeline AND prevent the job from getting dispatched. I haven't figured out a better way of getting the same behaviour. So, if you got any idea or concern regarding to the change, just ping me in Github and I am happy to discuss with you about it. Thank you so much for taking time reviewing it and I'm looking forward to hear from you!

### DIFF
--- a/lib/exq/middleware/pipeline.ex
+++ b/lib/exq/middleware/pipeline.ex
@@ -11,10 +11,18 @@ defmodule Exq.Middleware.Pipeline do
   * `event` - name of current middleware function, possible values are: `before_work`,
   `after_processed_work` and `after_failed_work`
   * `halted` - flag indicating whether pipeline was halted, defaults to `false`
+  * `terminated` - flag indicating whether worker and pipeline were halted, If
+      the flag was set to true, the job will not be dispatched and all after_*_work/1
+      will be skipped. For each specific middleware:
+      - Exq.Middleware.Job: Will NOT remove the backup from job queue
+      - Exq.Middleware.Logger: Will NOT record job as done or failed with timestamp
+      - Exq.Middleware.Manager: Will NOT update worker counter
+      - Exq.Middleware.Stats: Will NOT remove job from processes queue
   """
 
   defstruct assigns:      %{},
             halted:       false,
+            terminated:   false,
             worker_pid:   nil,
             event:        nil
 
@@ -32,6 +40,13 @@ defmodule Exq.Middleware.Pipeline do
   """
   def halt(%Pipeline{} = pipeline) do
     %{pipeline | halted: true}
+  end
+
+  @doc """
+  Sets `terminated` to true
+  """
+  def terminate(%Pipeline{} = pipeline) do
+    %{pipeline | terminated: true}
   end
 
   @doc """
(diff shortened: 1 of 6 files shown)


mix.lock
```diff
@@ -1,24 +1,24 @@
-%{"certifi": {:hex, :certifi, "0.3.0"},
-  "combine": {:hex, :combine, "0.7.0"},
-  "connection": {:hex, :connection, "1.0.2"},
+%{"certifi": {:hex, :certifi, "0.3.0", "389d4b126a47895fe96d65fcf8681f4d09eca1153dc2243ed6babad0aac1e763", [:rebar3], []},
```
akira: Can you clean up all this stuff in the mix file? Not sure why this was changed.

lib/exq/middleware/pipeline.ex
```diff
@@ -15,6 +15,7 @@ defmodule Exq.Middleware.Pipeline do
 
   defstruct assigns:      %{},
             halted:       false,
+            terminated:   false,
```
akira: Can you add documentation for this behavior on the moduledoc

akira:
I think approach is ok, if you can just clean up based on my comments it should be good

mix.lock
wli0503: Right. Will put the file in ```.gitignore```

lib/exq/middleware/pipeline.ex
wli0503: Definitely. I will be working on it. Thanks for pointing it out!

mix.lock
akira: I would leave .gitignore and just reset this particular file, as there's no need to update it unless mix.exs was also changed.

akira:
👍 Do you mind squashing and I can merge.

wli0503:
Squashed into a single commit and ready for merge. Thank you so much for taking time reviewing the change! Let me know if you have any other question or concern.

wli0503:
I noticed a travis-CI failure, specifically:
```
test test to_job_json performance (PerformanceTest)
     test/performance_test.exs:23
     ** (ArgumentError) argument error
```
Not sure what is going on with this one, from the first look it isn't related to my change and I can't reproduce this one on my local machine. @akira It would be great if you can take a look at it but I will take a look after work as well.

akira:
I think it was just a glitch.  Reran tests and merge.  Thanks!  👍

wli0503:
Thanks for merging it into core! I am really excited! One quick question, when will the package on ```Hex``` being updated so we can pull the new ```Exq``` with ```mix deps.get```?

akira:
@wli0503 I will update the package shortly, after verifying fixes for another issue.

wli0503:
@akira Got it. Thank you so much!

akira:
@wli0503 I published this on latest version, 0.7.2

wli0503:
@akira Acknowledged! Really appreciate for letting me know!

PR #203, opened 2016-07-29
upgrade timex dependency to ~> 3.0, use tzdata ~> 0.1.8

baron:
I upgraded the Timex dependency to 3.0. It's mostly minor renaming.

One issue I did encounter is that although all tests pass, using the latest version of tzdata, that runs as its own application would somehow cause gen server issues with the test suite.

Unfortunately, I have no clue with that, other than it is a common enough issue that tzdata comes in two flavors to handle the issue.


baron:
Not sure how to do it but I want this pull request deleted. I'll eventually delete my fork once the Timex dependency removal makes it into the main hex package.

akira:
@baron no problem, I will not merge this fork and can close it when we merge the other one in. Thanks for the change anyways 😄

akira:
Closed in favor of #204

PR #205, opened 2016-08-03
add job result and raw exception in pipeline

ananthakumaran:
This will provide interaction between worker and middleware. 

Usecase: Retry the job 5 times in normal circumstances, but also allow the worker to kill the job right away if some essential conditions are not met. This could be done easily by a middleware, if the worker return value is available in the pipeline.


akira:
👍 👍 Good idea!  Thanks for the PR!

PR #219, opened 2016-10-24
Expose retry via Exq.Api

chelsea:
To support https://github.com/akira/exq_ui/issues/5 I couldn't find any way in the current exposed API to retry a job immediately, so I had a go at adding the ability to immediately enqueue a retry, removing it from the retry queue.

I'm still pretty new to elixir so happy for feedback and to make adjustments.


akira:
Looks good!  Thanks! 👍

PR #229, opened 2017-01-14
Attempt and Unique jobs

danhawkins:
Myself and a friend has a go at a way to do unique jobs, I'm making this PR for feedback more than anything, so please let me know what you think

### DIFF
--- a/lib/exq/enqueuer/uniqueness.ex
+++ b/lib/exq/enqueuer/uniqueness.ex
@@ -1,73 +1,67 @@
 defmodule Exq.Enqueuer.Uniqueness do
+  @moduledoc """
+    This module is used to enforce job uniqueness when a job is first enqueued.
+    It does this by either receiving or generating a key and saving it in redis, (the "enqueue key").
+    Finally it saves a reference to this key referenced by the job number, (the "completion key").
+    From the middleware, we remove both keys when the job completes via the remove_locks/1 function.
+  """
+
   import Base, only: [encode16: 1]
   import String, only: [downcase: 1]
   alias Exq.Redis.Connection, as: Redis
   alias Exp.Middleware.Pipeline
 
-  # def with_unique_lock(callback, state, queue \\ "default", worker \\ "", args \\ []) do
-  #   with_unique_lock(callback, state, queue, worker, args, nil)
-  # end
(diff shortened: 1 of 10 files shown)


akira:
Thanks, I can take a look when I get a chance. Is your intent to keep this separate from Exq or to are you trying to merge this in?

If keeping it separate, you may want to keep the changes to the enqueuer outside - perhaps you can have your own own module that you use to wrap that can then use the unique logic?

akira:
I think overall it's a good approach. I like how this tracks the lock keys as well.  

I think as I mentioned above, my only issue is the modification of the enqueuer part of Exq. Without that part, this can be a completely separate hex package that someone can install and use on top of Exq. With the modification, you would need to keep merging upstream.  I think you can have the enqueue_unique parts be separate and still working within this structure?

danhawkins:
@akira thanks, I think your approach sounds much better. I'll try to decouple it

lib/exq/enqueuer/uniqueness.ex
```diff
@@ -0,0 +1,67 @@
...
+
+    if enqueue_key_exists?(redis, enqueue_key) do
+      {:job_not_unique, enqueue_key}
+    else
+      set_key(redis, enqueue_key)
+      {:reply, {:ok, jid}, _pipeline} = callback.()
+      completion_key = generate_completion_key(namespace, jid)
+      set_key(redis, completion_key, enqueue_key)
+      {:ok, jid}
+    end
```
ananthakumaran: https://redis.io/commands/setnx is the preferred way to handle this kind of scenario.

get # client a
get # client b
set # client a
set # client b

This will result in two jobs. Though it might not happen with one node as the requests are serialized by the gen server, but might happen if there are more than one nodes.

psteininger:
@danhawkins and @akira  is this feature still being developed? I would love to see it merged in and working.

akira:
@psteininger I don't believe this was created to be merged in. This should be a separate library from Exq providing this functionality.

matrinox:
Since @akira wants this to be a separate library, is there a link to the new repo? This would be a good feature for us to use

danhawkins:
Sorry I didn't get time to finish this and unlikely to for a while

PR #241, opened 2017-03-07
make backoff time calculation configurable

ananthakumaran:


lib/exq/support/config.ex
```diff
@@ -20,6 +22,7 @@ defmodule Exq.Support.Config do
     stats_batch_size: 2000,
     serializer: Exq.Serializers.JsonSerializer,
     node_identifier: Exq.NodeIdentifier.HostnameIdentifier,
+    backoff_time: {__MODULE__, :default_backoff_time},
```
akira: Can you use similar pattern as node_identifier, etc?  Either way is a valid approach, but I'd like it to be consistent.

akira:
Thanks!  👍 👊

PR #247, opened 2017-04-01
expose job metadata to worker process

### LINKED ISSUE #155: Job introspection from worker
Ability for worker to retrieve it's job metadata (for example, jid, retry count)

Perhaps support for this in worker:

```
 (job = %Job{}, param1, param2..)
```

Preferably, just provide an api like (suggested by edgurgel):

```
Exq.Jobs.mine(self) 
```

That would retrieve that data so that we don't have to change or provide multiple signatures for `perform` since this may not be as commonly used.

Suggested by olivermt on elixir-lang slack.

ananthakumaran:
Closes #155

### DIFF
--- a/lib/exq.ex
+++ b/lib/exq.ex
@@ -3,6 +3,7 @@ defmodule Exq do
   use Application
 
   import Exq.Support.Opts, only: [top_supervisor: 1]
+  alias Exq.Worker.Metadata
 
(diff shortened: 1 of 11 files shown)


akira:
👍 🔥 That's great!  Thanks for the PR!!

PR #256, opened 2017-04-17
address some variable unsafe mix compile warnings

joshk:
While using Exq, `mix deps.compile` gave three warnings on unsafe variable usage.

This PR addresses these warnings.

Below is the warning Mix would output:

```
warning: the variable "state" is unsafe as it has been set inside a case/cond/receive/if/&&/||. Please explicitly return the variable value instead. For example:

    case int do
      1 -> atom = :one
      2 -> atom = :two
    end

should be written as

    atom =
      case int do
        1 -> :one
        2 -> :two
      end
```


akira:
LGTM 👍   Thanks!

PR #264, opened 2017-05-05
Allow {:system, varname} format for configuration

stavro:
Would love to get Exq supporting the `{:system, "VARNAME"}` format for runtime config...

Thoughts?

-- EDIT --

If you don't like the `{:system, ""}` approach, there are others:

* Exq can specify a fallback environment variable for URL (not from config)
* Exq can provide an [on_init callback similar to Phoenix](https://github.com/phoenixframework/phoenix/blob/7fbf4fc0567a3276c978b434c4f0c658228af276/lib/phoenix/endpoint.ex#L63-L80) to update config when the application starts.
* Exq can [rely less on compile-time config and more on arguments](https://elixirforum.com/t/mix-config-evolutions/4423/19)

If you think you'd merge this, I can add additional tests for when the environment variable is an invalid value (which is why coveralls says the code coverage went down)

### DIFF
--- a/lib/exq/support/opts.ex
+++ b/lib/exq/support/opts.ex
@@ -1,5 +1,6 @@
 defmodule Exq.Support.Opts do
 
+  alias Exq.Support.Coercion
   alias Exq.Support.Config
 
   @doc """
@@ -27,8 +28,8 @@ defmodule Exq.Support.Opts do
       url
     else
       host = opts[:host] || Config.get(:host)
-      port = coerce_integer(opts[:port] || Config.get(:port))
-      database = coerce_integer(opts[:database] || Config.get(:database))
+      port = Coercion.to_integer(opts[:port] || Config.get(:port))
+      database = Coercion.to_integer(opts[:database] || Config.get(:database))
       password = opts[:password] || Config.get(:password)
       [host: host, port: port, database: database, password: password]
     end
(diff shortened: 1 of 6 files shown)


stavro:
@akira If you think you'd merge this, I can add additional tests for when the environment variable is an invalid value (which is why coveralls says the code coverage went down)

akira:
@stavro We can support `{:system, "VARNAME"}` format. 

I haven't looked at PR in detail yet, just curious, there is also hex package: https://github.com/Nebo15/confex/, does that simplify implementation?

stavro:
@akira I'm not sure that the logic is complex enough to warrant the addition of a dependency, but if you feel the additional configuration options granted by `confex` are worth it, I can certainly add it in.

lib/exq/support/opts.ex
```diff
@@ -42,6 +42,21 @@ defmodule Exq.Support.Opts do
     "#{name}.Redis.Client" |> String.to_atom
   end
 
+  defp coerce_integer(value) when is_integer(value) do
```
akira: Would it be better to move this somewhere in utils?  Maybe config?

akira:
@stavro looking at it, I think this PR as is will work. Maybe can add the additional test with invalid env variable. 

Thanks!

stavro:
PR updated - though coveralls is having some issues (I force pushed an update to try and get it to wake up again, but no luck).

akira:
Looks great, thanks for the PR Sean!!  👍  👏

PR #268, opened 2017-05-18
Add config tuple with default {:system, VAR, default}

kaelumania:


kaelumania:
The Travis CI issue with `Erlang/OTP Release 17.5 is not supported by Elixir 1.3.2` seems to be unrelated.

akira:
👍 thanks!

Issue #280, opened 2017-08-09
Test error on rand_bytes in 1.4.5


djKooks:
Hello.
I am keep facing on errors at running test. There are several faliures, but they are mostly caused by same reason.
```
test performance is in acceptable range (PerformanceTest)
     test/performance_test.exs:47
     ** (exit) exited in: GenServer.call(Exq, {:enqueue, "default", PerformanceTest.Worker, ["keep_on_trucking"], []}, 5000)
         ** (EXIT) time out
     stacktrace:
       (elixir) lib/gen_server.ex:737: GenServer.call/3
       test/performance_test.exs:54: anonymous fn/2 in PerformanceTest.test performance is in acceptable range/1
       (elixir) lib/enum.ex:1785: Enum.reduce_range_inc/4
       test/performance_test.exs:54: (test)

     The following output was logged:
     
     2017-08-10 00:17:00.433 [error]: GenServer Exq.Enqueuer terminating
     ** (UndefinedFunctionError) function :crypto.rand_bytes/1 is undefined or private. Did you mean one of:
     
           * rand_seed/0
           * rand_seed/1
     
         (crypto) :crypto.rand_bytes(16)
         lib/uuid.ex:281: UUID.uuid4/1
         lib/exq/redis/job_queue.ex:363: Exq.Redis.JobQueue.to_job_serialized/5
```
It looks like problem from unsupporting `crypto.rand_bytes` anymore. Does anyone has same problem with me?

My elixir version is 1.4.5.

akira:
This looks like an issue with uuid package. I wonder if pinning a higher version of that would help this.

djKooks:
@akira Thanks. It works well after bump up uuid version. But I'm not sure if this will be fine on older Elixir version. Would you look on?

PR #281, opened 2017-08-11
Update version of uuid dependency

djKooks:
I bump up uuid version to solve issue #280 . It works fine on Elixir 1.4.5, but not been tested on older version.
Please look on. Thanks.


akira:
👍 Tests pass and use older version, so looks good. Thanks!

PR #282, opened 2017-08-15
add whereis/1 as shortcut for pid of exq instance

techgaun:
I think having Exq.whereis/1 as a separate function is useful. I am using it right now.


akira:
👍 sounds good

PR #285, opened 2017-09-05
Pass socket_opts to redix

matrinox:
This paves the way for SSL support but in the meantime allows one to configure the socket options for the TCP connection

NOTE: the upgrade of redix to v0.6.1 is not important and I can revert it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ elixir:
   - 1.3.2
   - 1.4.0
 otp_release:
-  - 17.5
   - 18.0
   - 19.0
 before_script:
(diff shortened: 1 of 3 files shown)


akira:
Sorry, been traveling and hadn't had a chance to look, will take a look shortly.

matrinox:
@akira would you also accept a patch for an older version, like 0.7.3? I need to add ssl support to another app using elixir 1.2 before it's migrated to 1.3. I can always just use the fork but would be nice.

akira:
Looks good but let's revert the redix upgrade if it's not necessary.

matrinox:
@akira all done

akira:
👍 Thanks!

Issue #261, opened 2017-04-29
Requeue failed jobs when a node crashes and does not come back online.


lpil:
Currently requeuing of jobs after a crash is the responsibility of a new node with the same node ID as the crashed node. The system would be more resilient to failure (and it would be easier to resize the cluster) if there was a mechanism for other nodes to perform this clean up.

## A possible approach

I've little experience in this kind of systems, but I was thinking something along these lines could work:

1. Nodes heartbeat to Redis periodically. This is recorded using their node id and a timestamp using Redis local time.
    - A set sorted by timestamp could be employed here to allow more efficient time based querying.
    - Heartbeats could be piggybacked onto the existing polling for jobs to prevent additional network overhead. Lua scripting may be required here.
2.  Nodes periodically check to see if any members of the set have not registered a heartbeat within a configurable set timeout. If a node is found to be missing they can requeue their jobs and deregister the node from the records in Redis.

There are problems with this approach:

If a node is isolated from Redis and no longer able to heartbeat (say, during a network partition) a node may be falsely assumed to be dead, causing the jobs in progress when the partition started to be requeued even though they finished successfully. Depending on your workload this may not be acceptable.

- Idempotent jobs can prevent this from causing damage, but would still cost worker time.
- Using a longer heartbeat timeout before requeuing can reduce the chance of this happening.

### Implementation

If we implement the heartbeat (step 1) in Exq the second part can be implemented as an additional library that depends on Exq. This gives users a way to opt into this behaviour, and helps Exq avoid having to manage deprecating this feature if it is decided that it is not wanted moving forward.

I don't believe that this library would need to share any configuration or code with Exq other than the address of the Redis instance being used and the Redis namespace.

It would be beneficial for the protocol that Exq uses to be documented and declared stable enough for this library to also use (as it will be modifying the same keys that Exq modifies).

---

Would be great to hear your thoughts. Even if this proposed approach isn't right I think this issue is one that would be good to try and solve.

Cheers,
Louis

akira:
+@joshk, we spoke about this. 

1. Having a node level heartbeat would work well for this. I think this is what would be needed to allow other nodes to take over when a node disappears. 

2. As you mentioned, nodes can check periodically if any heartbeats are stale in sorted set, and requeue job if so.

Job should be idempotent, so I don't think we should attempt to resolve the issue with disconnected nodes. The heartbeat timeout can be configurable.  

[...]

PR #289, opened 2017-10-11
Update Manager and stats to correctly store information about proceses

ondrejbartas:
This fixes problem with Sidekiq Web UI which has changed API from version 4.

Now we correctly add data about manager and processes to redis

### DIFF
--- a/lib/exq/manager/heartbeat_server.ex
+++ b/lib/exq/manager/heartbeat_server.ex
@@ -21,7 +21,7 @@ defmodule Exq.Heartbeat.Server do
       queues: opts[:queues],
       poll_timeout: opts[:poll_timeout]
     }
-    schedule_work(state, true, 0)
+    schedule_work(state, true)
     {:ok, state}
   end
 
@@ -30,7 +30,7 @@ defmodule Exq.Heartbeat.Server do
     master_data = Map.from_struct(master_state)
     current_state = %{struct(State, master_data) | name: state.name}
     init_data = if status, do: [["DEL", "#{current_state.name}:workers"]], else: []
-    data = init_data ++ JobStat.get_redis_commands(
+    data = init_data ++ JobStat.status_process_commands(
       current_state.namespace,
       current_state.node_id,
       current_state.started_at,
@@ -46,7 +46,7 @@ defmodule Exq.Heartbeat.Server do
     {:noreply, current_state}
   end
 
-  defp schedule_work(state, status \\ false, timeout \\ 1000) do
+  defp schedule_work(state, status \\ false) do
     Process.send_after(state.name, {:get_state, self(), status}, 1000)
   end
 end
--- a/lib/exq/manager/heartbeat_server.ex
+++ b/lib/exq/manager/heartbeat_server.ex
@@ -1,61 +1,55 @@
-defmodule HeartbeatServer do
+defmodule Exq.Heartbeat.Server do
   use GenServer
   alias Exq.Redis.JobQueue
   alias Exq.Redis.Connection
+  alias Exq.Support.Config
   alias Exq.Support.Time
+  alias Exq.Redis.JobStat
 
-  def start_link(master_state) do
-    GenServer.start_link(__MODULE__, master_state)
+  defmodule State do
+    defstruct name: nil, node_id: nil, namespace: nil, started_at: nil, pid: nil, queues: nil, poll_timeout: nil, work_table: nil, redis: nil
   end
 
-  def init(master_state) do
-    worker_init = [ ["DEL", "#{redis_worker_name(master_state)}:workers"] ]
-      ++ getRedisCommands(master_state)
-    Connection.qp!(master_state.redis, worker_init)
-
-    schedule_work()
-
-    {:ok, master_state}
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts)
   end
 
-  def handle_info(:heartbeat, master_state) do
-    schedule_work()
+  def init(opts) do
+    state = %State{
+      name: Exq.Manager.Server.server_name(opts[:name]),
+      redis: opts[:redis],
+      node_id:  Config.node_identifier.node_id(),
+      namespace: opts[:namespace],
+      queues: opts[:queues],
+      poll_timeout: opts[:poll_timeout]
+    }
+    schedule_work(state)
+    {:ok, state}
+  end
 
-    current_state = GenServer.call(Map.get(master_state, :pid), :get_state)
-    Connection.qp!(current_state.redis, getRedisCommands(current_state))
-
+  def handle_cast({:heartbeat, master_state}, state) do
+    schedule_work(state)
+    current_state = struct(State, Map.from_struct(master_state))
+    Connection.qp!(
+      current_state.redis,
+      JobStat.get_redis_commands(
+        current_state.namespace,
+        current_state.node_id,
+        current_state.started_at,
+        current_state.pid,
+        current_state.queues,
+        current_state.work_table,
+        current_state.poll_timeout
+        )
+      )
     {:noreply, current_state}
   end
 
-  defp schedule_work() do
-    Process.send_after(self(), :heartbeat, 1000)
+  defp schedule_work(state) do
+    Process.send_after(state.name, {:get_state, self()}, 1000)
   end
 
   defp redis_worker_name(state) do
     JobQueue.full_key(state.namespace, "#{state.node_id}:elixir")
   end
-
-  defp getRedisCommands(state) do
-    name = redis_worker_name(state)
-    [
-      ["SADD", JobQueue.full_key(state.namespace, "processes"), name],
-      ["HSET", name, "quiet", "false"],
-      ["HSET", name, "info", Poison.encode!(%{ hostname: state.node_id, started_at: state.started_at, pid: "#{:erlang.pid_to_list(state.pid)}", concurrency: cocurency_count(state), queues: state.queues})],
-      ["HSET", name, "beat", Time.unix_seconds],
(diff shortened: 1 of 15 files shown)


ondrejbartas:
It is still not working, problem with cleaning up processes after init :(

akira:
Thanks! Will need to take a look. 

Also, I'm assuming this will require changes to [exq_ui](github.com/akira/exq_ui)?

ondrejbartas:
@akira yes exq_ui will need to change.

lib/exq/manager/server.ex
```diff
@@ -160,13 +161,37 @@ defmodule Exq.Manager.Server do
...
+      ["HSET", name, "info", Poison.encode!(%{ hostname: state.node_id, started_at: state.started_at, pid: "#{:erlang.pid_to_list(state.pid)}", concurrency: cocurency_count(state), queues: state.queues})],
+      ["HSET", name, "beat", Time.unix_seconds],
+      ["EXPIRE", name, (state.poll_timeout / 1000 + 5)],
+    ]
+    Connection.qp!(state.redis, worker_init)
+
     {:ok, state, 0}
   end
 
+  def cocurency_count(state) do
```
akira: Spelling

lib/exq/manager/server.ex
```diff
@@ -251,6 +281,17 @@ defmodule Exq.Manager.Server do
 
       job_results = jobs |> Enum.map(fn(potential_job) -> dispatch_job(state, potential_job) end)
 
+      # Update worker info in redis that it is alive
```
akira: So this manages worker state each time it dequeues, is this how Sidekiq does it?

akira: This seems to be adding overhead.  Looking at performance test:

without lines 284:293
`2017-10-29 13:25:34.217 [debug]: Perf test took 0.680126 secs`

with lines 284:293
`2017-10-29 13:23:03.301 [debug]: Perf test took 1.000115 secs`

(To run this I set `ExUnit.start(capture_log: false)` in test_helper.exs)

akira:
Looking at the code again, I think the heartbeat mechanism could be tweaked. It seems like some overhead to update this on every dequeue - when it can be done a lot less frequently. Perhaps we can have a separate GenServer that will do the heartbeat?  

This PR actually goes makes some progress on issue #261.  For the full implementation, a random uuid can be generated whenever a new heartbeat worker comes online. If the system crashes, this should crash as well. The cleanup process can then look through uuids and see whichever ones are stale, and re-queue those message.  However, even if all this is not implemented for this PR, maybe we can start with the separate GenServer that simply starts up, inserts worker information, and then heartbeats - and will crash when system crashes?

Let me know your thoughts.

lib/exq/manager/server.ex
ondrejbartas: Hi @akira, biggest problem is cleanup of queues which happens parallel by calling: https://github.com/akira/exq/pull/289/files/adbd0d7657abff63dcdfdb1db5a7240c7ce4d4a6#diff-ee19dd3b88a278d7cf31b33f2b2b5705R147

I am new to elixir so I have no experience how to do it in one run, first, cleanup, then whole code.

ondrejbartas: then you wil need just: 
```
name = redis_worker_name(state)
      worker_init = [
         ["HSET", name, "beat", Time.unix_seconds],
         ["EXPIRE", name, (state.poll_timeout / 1000 + 5)], # expire information about live worker in poll_interval + 5s
       ]
Connection.qp!(state.redis, worker_init)
```

ondrejbartas: also this code can run every second and result will be same

akira: Yes - for Sidekiq, seems like it is running this in a separate thread:
https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/launcher.rb#L25
That is why I was thinking if we can just run this separately as well, would it be possible to do this in a separate GenServer?  

If you need to do synchronous methods from the Manager to the GenServer, you can still do GenServer.call instead of GenServer.cast and it will wait for the call. Let me know if you think this is doable or not.  The other option is perhaps lower the frequency of this call so it doesn't do it in every dequeue.   

The other thing I noticed is the section with HSET, etc is duplicated - it would be nice to take this out into a function at least, or function in JobStat module.

TondaHack:
Hi @akira, 

Could you check these changes we made? 

Thanks

ondrejbartas:
@akira Hi can you review this PR again?

akira:
Sorry, have been slammed at work lately.  Will try to take a look when things lighten up.  Thanks!

lib/exq/manager/server.ex
```diff
@@ -160,13 +156,25 @@ defmodule Exq.Manager.Server do
...
                    }
 
     check_redis_connection(opts)
+
+    # Cleanup stale stats
+    rescue_timeout(fn ->
+      Exq.Stats.Server.cleanup_host_stats(state.stats, state.namespace, state.node_id, state.pid)
+    end)
+
+    HeartbeatServer.start_link(state)
```
akira: Starting a process this way is not recommended. The proper way of starting it is to add this to the supervision tree:  https://github.com/akira/exq/blob/master/lib/exq/support/mode.ex#L26

lib/exq/manager/heartbeat_server.ex
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  defp redis_worker_name(state) do
+    JobQueue.full_key(state.namespace, "#{state.node_id}:elixir")
+  end
+
+  defp getRedisCommands(state) do
+    name = redis_worker_name(state)
+    [
+      ["SADD", JobQueue.full_key(state.namespace, "processes"), name],
```
akira: Can we refactor redis commands to be in one of the other modules? For example. RedisStat. 

Also, nit, using camel case for methods.

lib/exq/redis/job_stat.ex
```diff
@@ -145,4 +186,30 @@ defmodule Exq.Redis.JobStat do
         val
     end
   end
+
+  def get_redis_commands(namespace, node_id, started_at, master_pid, queues, work_table, poll_timeout) do
```
akira: Maybe name this something more descriptive, above methods that are similar are named something like <usage>_commands

lib/exq/redis/job_stat.ex
```diff
@@ -145,4 +186,30 @@ defmodule Exq.Redis.JobStat do
...
+      ["HSET", name, "beat", Time.unix_seconds],
+      ["EXPIRE", name, (poll_timeout / 1000 + 5)]
+    ]
+  end
+
+  defp redis_worker_name(namespace, node_id) do
+    JobQueue.full_key(namespace, "#{node_id}:elixir")
+  end
+
+  defp cocurency_count(queues, work_table) do
```
akira: This is mispelled.

akira:
Looks like build is failing from some compilation issues

TondaHack:
Hi @akira, 

Thanks for the review. A few of builds failed because of compilation issue, but the rest of them because of tests. I know that reason is Exq.Heartbeat.Server. I am trying to find the issue but every build shows different failing tests. It's happening in `/exq/test/exq_test.exs` and with tests related to `ExqTest.PerformWorker`. 

I've tried to run the first heartbeat with smaller or no a timeout. It didn't work. Also, I tried to extend assert_receive timeout from 100 to 200 or 400 ms. There were less failing tests but still some. Could you take a look, please?

akira:
@TondaHack I see the tests pass now, looks like you figured out the issue?

TondaHack:
@akira Yes, I think so. The clean up of Heartbeat server helped. What do you think about current changes?

akira:
@TondaHack @ananthakumaran I wonder if it would be better to have a generated node ID instead of just using the host. If you look at #321 some people may be running multiple workers on the same machine (at least this case is it true during deploys). However, this obscures each node heartbeat since two nodes will be registering the same heartbeat. 

If we use a more ephemeral key for a node, for example PID or UUID (plus the hostname), it would be unique even if multiple processes are running for the same node. 

The other comment I had is, this is a big change to the current system and may break some compatibility. I wonder if there's a way to deploy this through some config flag, where we can keep the old method, and allow a user to switch to use this new method. This would be a much lower risk way of deploying this change without impacting anyone.

leeicmobile:
@akira any chance we can just make the heartbeat solution a major version release rather than a config change currently a potential blocker in my production deployment to k8s cluster. #imselfish

akira:
> @akira any chance we can just make the heartbeat solution a major version release rather than a config change currently a potential blocker in my production deployment to k8s cluster. #imselfish

I have been very busy lately with work / life and have not had much time lately. I would consider merging this if following happened:

1. Someone can review this diff. 
2. This can be tested in a running ensure no new bugs have been introduced. 
3. ExqUI is updated to understand this format (it is not backwards compatible with this PR). 
4. (Nice to have) Add dynamic UUID to node identifier, so it handles restarts on the same host.  

Let me know if anyone has done this or can spend some time doing any of these.  It is a big change and I just didn't want to merge it without proper due diligence.

ananthakumaran:
@leeicmobile There is an open [PR](https://github.com/akira/exq/pull/392) which implements the heartbeat. If you are still waiting for it you could help with testing it.

ananthakumaran:
> 1. Someone can review this diff.
> 2. This can be tested in a running ensure no new bugs have been introduced.
> 3. ExqUI is updated to understand this format (it is not backwards compatible with this PR).
> 4. (Nice to have) Add dynamic UUID to node identifier, so it handles restarts on the same host.

I would love to get this merged into master. I can help with 1 & 2. I believe 4 is already fixed in master. I am not sure about 3, as I am not familiar with exq_ui codebase and we don't use it as well.

ananthakumaran:
fixed in master

PR #302, opened 2018-02-08
Ensure sidekiq jobs are retried.

deepfryed:
Default option in sidekiq sets retry => true [1]. Adding a matcher in exq to retry such jobs.

[1] https://github.com/mperham/sidekiq/blob/master/lib/sidekiq.rb#L41

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -17,7 +17,7 @@ config :exq,
   redis_timeout: 5000,
   genserver_timeout: 5000,
   test_with_local_redis: true,
-  max_retries: 0,
+  max_retries: 1,
   stats_flush_interval: 5,
   stats_batch_size: 1,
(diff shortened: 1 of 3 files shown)


config/test.exs
```diff
@@ -17,7 +17,7 @@ config :exq,
   redis_timeout: 5000,
   genserver_timeout: 5000,
   test_with_local_redis: true,
-  max_retries: 0,
+  max_retries: 1,
```
akira: Rather than changing config value for all tests, can we just change for your specific test?

deepfryed: Done

akira:
👍 Thanks for the PR!

deepfryed:
cheers, please give me a shout out when you bump version on hex.pm.  thanks!

akira:
@deepfryed just bumped

deepfryed:
thanks 🚀

PR #309, opened 2018-04-01
Add an api to list all subscriptions(active queues)

robobakery:
It would be great if i can access to the list of all subscriptions via api.

I've added this functionality for my usecase of showing all active queues.

example usage:
{:ok, subscriptions} = Exq.subscriptions(Exq)
{:ok, ["two", "four", "five"]}

short story:
I have 5 queues named one, two, three, four, five
and i want to highlight active queue which is currently subscribed.
if i am subscribing two, four, five
result would be like this: one, `two`, three, `four`, `five`

### DIFF
--- a/lib/exq.ex
+++ b/lib/exq.ex
@@ -46,6 +46,14 @@ defmodule Exq do
   end
 
   @doc """
+  List all subscriptions(active queues)
+    * `pid` - PID for Exq Manager or Enqueuer to handle this
+  """
+  def subscriptions(pid) do
+    GenServer.call(pid, :subscriptions)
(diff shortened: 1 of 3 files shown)


robobakery:
thanks for @mbuhot from slack.

`:sys.get_state(Exq).queues` will do the same thing.

akira:
Thanks, looks good! Can we add a test?

robobakery:
i've added some tests :)
please check it out and advice me if any modification needed.

akira:
LGTM!  🚢   Thanks for the additions!

Issue #311, opened 2018-04-05
Exq does not wait for long running job to terminate


frahugo:
We configured Exq with `shutdown_timeout: 30_000` and Exq still terminates within 5 seconds.

PR #312, opened 2018-04-05
Have top supervisor wait for worker drainer to gracefully terminate

frahugo:
By default, the supervisor waits only 5 seconds.  It should wait for as long as the :shutdown_timeout option is.

Fixes issue #311

### DIFF
--- a/test/exq_test.exs
+++ b/test/exq_test.exs
@@ -362,6 +362,20 @@ defmodule ExqTest do
     assert_received {"short"}
   end
 
+  @tag :slow
+  test "configure worker shutdown time greater than 5000ms" do
+    Process.register(self(), :exqtest)
+    {:ok, sup} = Exq.start_link([shutdown_timeout: 6500])
+    {:ok, _} = Exq.enqueue(Exq, "default", ExqTest.SleepWorker, [7000, :long])
+    {:ok, _} = Exq.enqueue(Exq, "default", ExqTest.SleepWorker, [6000, :short])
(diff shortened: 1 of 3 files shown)


test/exq_test.exs
```diff
@@ -362,6 +362,20 @@ defmodule ExqTest do
     assert_received {"short"}
   end
 
+  @tag :slow
+  test "configure worker shutdown time greater than 5000ms" do
+    Process.register(self(), :exqtest)
+    {:ok, sup} = Exq.start_link([shutdown_timeout: 6500])
+    {:ok, _} = Exq.enqueue(Exq, "default", ExqTest.SleepWorker, [7000, :long])
```
akira: Looks good, can we test smaller amount with this and just as long as `refute_received {"long"}` works it would still be a valid check?

frahugo: The default supervisor shutdown value is 5sec in OTP if not provided in the child spec.  Should we instead check that the child spec for the WorkerDrainer has the right shutdown value?

akira: I see, the only concern I had is adding extra time to the tests (10sec vs 16sec) - if that is an easy change, we can check the WorkerDrain instead.

frahugo: Ok sounds good. I will add a test for the Mode module and assert the child spec has the shutdown option.

akira:
Looks good 👍  Thanks for the PR!

PR #315, opened 2018-04-26
Fix memory leak with scheduler

kareemk:
We payloads that can be up to 400kb in size and when there is a flood of errors that continuously failed we experienced acute memory bloat, seeing some processes grow to 4GB+. 

Investigating the issue it looks like the issue is that the scheduler re-enqueues jobs by fetching *all* schedulable jobs. This fix limits this fetch to a configurable page size.

The ideal solution is to use a LUA script which moves the jobs from the schedule sorted set to the queue atomically and without the round-trip. I plan on implementing this soon and adding it to this pull request.

### DIFF
--- a/lib/exq/redis/job_queue.ex
+++ b/lib/exq/redis/job_queue.ex
@@ -125,11 +125,11 @@ defmodule Exq.Redis.JobQueue do
   end
 
   def scheduler_dequeue(redis, namespace) do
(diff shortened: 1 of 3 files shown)


akira:
Also - how did you come up with default value of 10?  Seems that a higher value could also potentially work without impacting memory.

kareemk:
We're still seeing memory leak issues with this fix deployed, so I'm going to go ahead and implement the LUA approach

PR #317, opened 2018-05-04
trim dead jobs queue

ananthakumaran:
We had a high redis memory usage problem and after some investigation it turned out that there were more than 500K dead jobs. This PR addresses this problem by trimming based on two configs (size and timeout). Sidkiq provides the same [config](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq.rb#L34-L35) and the [implementation](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/api.rb#L661-L665) is more or less the same

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,19 @@ config :exq,
 
 Note that ```scheduler_enable``` has to be set to ```true``` and ```max_retries``` should be greater than ```0```.
 
+### Dead Jobs:
+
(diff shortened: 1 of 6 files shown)


ananthakumaran:
I would say probably not. This leak occurs in Redis, whereas #315 seems to happen in the worker. Our worker memory usage never went beyond 500Mb.

PR #318, opened 2018-05-13
Dequeue & Re-enqueue using LUA script

kareemk:
This supports dequeuing and re-enqueuing to the scheduled queues using an atomic LUA script as mentioned in https://github.com/akira/exq/pull/315. 

This also stored JID strings in the scheduled queues and the jobs and associated metadata in Hashes to support overriding the JID to support uniquely scheduling jobs using an idempotency key. 

Note, we've seen that a default page size of 10 is sufficient to clear 1000s of jobs very quickly and the benefit of a smaller page size with this implementation is keeping LUA script execution time small given Redis's single threaded architecture.

### DIFF
--- a/lib/exq/redis/script.ex
+++ b/lib/exq/redis/script.ex
@@ -1,0 +1,79 @@
+defmodule Exq.Redis.Script do
+  alias Exq.Redis.Connection
+
+  defmodule Prepare do
+    def script(source) do
+      hash = :crypto.hash(:sha, source)
+                    |> Base.encode16(case: :lower)
+
+      {hash, source}
+    end
+  end
+
+  @scripts %{
+    schedule: Prepare.script("""
+      local schedule_queue_key, job_queue_key = KEYS[1], KEYS[2]
+      local jid, score, job_serialized = ARGV[1], ARGV[2], ARGV[3]
+
+      redis.call('ZADD', schedule_queue_key, score, jid)
+      redis.call('HSET', schedule_queue_key .. ':job', jid, job_serialized)
+      redis.call('HSET', schedule_queue_key .. ':job_queue', jid, job_queue_key)
+      """),
+    schedule_dequeue: Prepare.script("""
+      local schedule_queue_key, queues_key = KEYS[1], KEYS[2]
+      local score = ARGV[1]
+
+      local dequeue_count = 0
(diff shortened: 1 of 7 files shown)


lib/exq/redis/script.ex
```diff
@@ -0,0 +1,79 @@
...
+      for i, jid in ipairs(redis.call('ZRANGEBYSCORE', schedule_queue_key, 0, score, 'LIMIT', 0, page_size)) do
+        local job_serialized = redis.call('HGET', schedule_queue_key .. ':job', jid)
+        local job_queue_key = redis.call('HGET', schedule_queue_key .. ':job_queue', jid)
+
+        redis.call('SADD', 'queues', queues_key, job_queue_key)
+        redis.call('LPUSH', job_queue_key, job_serialized)
+
+        redis.call('ZREM', schedule_queue_key, jid)
+        redis.call('HDEL', schedule_queue_key .. ':job', jid)
+        redis.call('HDEL', schedule_queue_key .. ':job_queue', jid)
```
ananthakumaran: It looks like the data structure used to store the schedules has changed. How would the upgrade work for existing users? I am assuming the changes are required because queue_name is not available in lua context. Is it possible to parse the job using [cjson](https://redis.io/commands/eval#cjson) or that would be too slow?

seanwash:
I'm running into the memory leak as outlined in #315 as well. Is there anything I can do to assist in geetting this out the door?

kareemk:
Do you have high error rates by any chance?

seanwash:
@kareemk for one particular job we can. We're working with the elixir elasticsearch lib which doesn't support bulk index updating yet (it's in progress), so our ES provider will sometimes rate limit some reindex jobs if they happen too quickly.

kareemk:
Ok, we ended up going with changes that are on [this branch](https://github.com/itskoko/exq/tree/scheduler-memory-leak) - the biggest contributor was Stats which we added a config setting to disable

seanwash:
@kareemk Cool, I'll take a look there. Thanks!

akira:
@kareemk @seanwash There were some comments with this that were not responded to, also I think there are some merge conflicts currently.

akira:
@kareemk @seanwash bumping this up.  If you guys have time to reply to comments / rebase for potentially merging.. Thanks!

PR #325, opened 2018-07-06
Update UUID dep in mix.exs

tzilist:
the UUID lib has changed it's name to `elixir_uuid`. The API is exactly the same. Having the old `uuid` and `elixir_uuid` deps is causing dep clash issues in that they both occupy the same namespace. This PR will bring it in line with the newest version and prevent future namespace clashes.

See the note here for more details:
https://github.com/zyro/elixir-uuid#elixir-uuid

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -34,7 +34,7 @@ defmodule Exq.Mixfile do
   # { :foobar, "0.1", git: "https://github.com/elixir-lang/foobar.git" }
   defp deps do
     [
-      { :elixir_uuid, ">= 1.2"},
+      { :elixir_uuid, ">= 1.2.0"},
       { :redix, ">= 0.5.0"},
       { :poison, ">= 1.2.0 or ~> 2.0"},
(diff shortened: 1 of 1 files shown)


doughsay:
This is causing a bit of a mess.  If you have a project that uses multiple dependencies that depend on uuid, you have to have ALL those dependencies update to the new name, otherwise you get conflicts.  So, as a library maintainer, it's annoying to be the first to update, because then you break everyone else's projects, but if you're late to update, people come banging on your door.

Hah, well in any-case, please update because others have already begun the change!

tzilist:
Agreed, the library change has caused all sorts of issues for us :(

akira:
👍 thanks for the PR.

pacoguzman:
shouldn't we update this too? https://github.com/tzilist/exq/blob/dd03b3ca5dd60f75fb972b11a6ee7130606ca442/mix.exs#L29

akira:
@pacoguzman good catch, just fixed

PR #327, opened 2018-07-14
Fix init func warning

kanmo:
Fix below warning.

```
warning: function init/1 required by behaviour GenServer is not implemented (in module Exq.Middleware.Server).

We will inject a default implementation for now:

    def init(args) do
      {:ok, args}
    end

You can copy the implementation above or define your own that converts the arguments given to GenServer.start_link/3 to the server state.

  lib/exq/middleware/server.ex:1
```


akira:
Thanks 👍

PR #331, opened 2018-08-01
Support load from runtime env for more config params

LysanderGG:
First of all, thank you for the great library.

When using it I was surprised that some of the config parameters could use the syntax  `{:system, VAR}` and some others could not.

This MR adds `{:system, VAR}` format support for integer parameters that are not supporting it yet and to boolean parameters.

In addition to integers, I supported `:infinity` for the `:concurrency` parameter. I assumed that no other parameter could be `:infinity` as there are no such tests or documentation.

I can add support for list of string in this or another MR (for `:queues`) if wanted.
A proposition of format could be splitting on comas, i.e. `"a,b,c"` parsed as `["a", "b", "c"]`.

Any suggestion is welcome.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -16,6 +16,7 @@ config :exq,
   poll_timeout: 100,
   redis_timeout: 5000,
   genserver_timeout: 5000,
+  shutdown_timeout: 5000,
   reconnect_on_sleep: 100,
   dead_max_jobs: 10_000,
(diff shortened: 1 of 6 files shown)


akira:
Sure, thanks for PR!  👍   Commas could work for array params if needed.

akira:
@LysanderGG Actually it looks like the build failed?

LysanderGG:
@akira I fixed it

akira:
👍 Thanks!

PR #335, opened 2018-09-18
Remove Elixir 1.7 warning for Enum.chunk

bmanuel:
This will remove the warning about Enum.chunk/2 while preserving functionality for users with Elixir 1.4 and below.

Warning:
```
warning: Enum.chunk/2 is deprecated. Use Enum.chunk_every/2 instead
  lib/exq/redis/job_queue.ex:285
```


ananthakumaran:
Thanks for the PR, this has been fixed in the master now.

PR #338, opened 2018-09-26
Specify less than 0.8.0 on redix version in mix.exs

### LINKED ISSUE #336: Redix.start_link/2 removed in 0.8.0 and returned but deprecate in 0.8.1
# To Reproduce
Change redix version to ">= 0.8.0" and run tests.

# Problem
A [breaking change in Redix v0.8.0](https://github.com/whatyouhide/redix/blob/master/CHANGELOG.md#v080) unified the redis and connection options in start_link.  This causes a `no function clause matching in Redix.start_link/2` exception

### LINKED ISSUE #337: Redix.start_link fails with "unknown option: :backoff"
Redix 0.8.1 does not accept the "backoff" configuration parameter, so exq startup fails.

The latest version of Redix that supports the backoff option is 0.7.1.

buob:
There were some breaking changes introduced in redix 0.8.0

Closes #336
Closes #337

(btw, does `mix deps.get` not look at the mix.lock of the dependency to get the sub-dependencies?  I would have expected that but mine got redix 0.8.1 so it failed ¯\_(ツ)_/¯)

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -35,7 +35,7 @@ defmodule Exq.Mixfile do
   defp deps do
     [
       { :elixir_uuid, ">= 1.2.0"},
-      { :redix, ">= 0.5.0"},
+      { :redix, ">= 0.5.0 and < 0.8.0"},
       { :poison, ">= 1.2.0 or ~> 2.0"},
       { :excoveralls, "~> 0.6", only: :test },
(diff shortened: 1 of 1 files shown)


akira:
No it does not pull from lock file.  Thanks!

PR #344, opened 2018-10-11
Replace unsupported redix options

ryansch:
## Backoff

The `backoff` option was removed from Redix in 2016: https://github.com/whatyouhide/redix/commit/f702a4f915dcc7a604d784e4f89b2828be505ce4#diff-b92857fa12652fb63155ed88e9f02b03

This PR aligns exq's parameter naming with redix's.

See: https://hexdocs.pm/redix/reconnections.html

## Timeout

The `timeout` option is likewise not supported by redix.  There is a `sync_connect` option that controls how connection issues are treated.  Docs: https://hexdocs.pm/redix/reconnections.html#synchronous-or-asynchronous-connection

PR #345, opened 2018-10-12
Generalize redis options as a keyword list

ryansch:
Closes #344 

### Backoff

The `backoff` option was removed from Redix in 2016: https://github.com/whatyouhide/redix/commit/f702a4f915dcc7a604d784e4f89b2828be505ce4#diff-b92857fa12652fb63155ed88e9f02b03

See: https://hexdocs.pm/redix/reconnections.html

### Timeout

The `timeout` option is likewise not supported by redix.

## Design Motivation

After submitting #344 I tried to set up ssl on redix 0.8.x and once again found myself missing redix options.  Exq shouldn't really care what options I want to pass to redix and should just do it.  I decided to directly expose the keyword list that redix is expecting.


akira:
Thanks for the PR!  So this would break some backwards compatibility with the previous settings? Is worth passing / translating the old settings to the redis_options?

ryansch:
@akira The old settings have been silently ignored for quite some time and don't really have a new setting that we could map them to.

akira:
Should this PR remove the Redix version limitation?

https://github.com/akira/exq/blob/master/mix.exs#L38

Also, does this fix #336 and #337 ?

jerryluk:
Would be good if it supports redix > 0.8.0

ryansch:
Sorry this one slipped through the cracks.

@akira This does fix #336 and #337.  I'm running this in production with the following additional patch: https://github.com/outstand/exq/commit/b2dc6c66436dd00cd5d0bbc7dc73b7cbe4e205f5

sbennett33:
Any movement on this? This PR does appear to remove the Redix version limitation. I'm also currently running the patched branch referenced by @ryansch and not seeing any issues.

akira:
Yes I was planning on merging soon just hadn’t gotten a chance.  @ryansch
if you get a chance can you include the patch on this branch as well?

ryansch:
@akira Done!

akira:
Thanks @ryansch!  Seems all the tests are failing. I think some of them would have to be changed as there are some mocking behavior being done, others I'm not sure of as it seems related to other missing packages.

ananthakumaran:
Redix has added support for sentinel in 0.9.0, so if we are ok with breaking compatibility, we can remove support for `redis_worker` option. It's one of the tests which is failing.

PR #346, opened 2018-10-12
Don't assume redis_opts is enumerable

ryansch:
When the redis configuration is just the url string and there's a connection error, exq raises an exception instead of displaying the error message.


akira:
👍 thanks!

PR #351, opened 2018-12-06
Use mix format

chulkilee:
Use the mix formatter.


akira:
@chulkilee Thanks for the PR and cleanup!!  👍 ✨

PR #356, opened 2019-01-14
Fix Elixir deprecation change to singular :microsecond

KalvinHom:


ananthakumaran:
Just to confirm, this is backward compatible?

dmarkow:
@ananthakumaran All releases since Elixir 1.4 (about 2 years ago) normalized both `:microsecond` and `:microseconds` to whatever Erlang expects. It's only Elixir 1.8 that started warning about using the plural form (so I assume 1.9 will fully raise an error instead of a warning). So it won't be backward compatible with Elixir 1.3 and earlier, but the `DateTime` module didn't exist then either so Exq is already not compatible with it.

[Elixir 1.4 normalized it](https://github.com/elixir-lang/elixir/blob/v1.4/lib/elixir/lib/system.ex#L811) to `:micro_seconds` and [Elixir 1.8 keeps it](https://github.com/elixir-lang/elixir/blob/v1.8/lib/elixir/lib/system.ex#L888) as `:microsecond` (I think Erlang's expected arguments changed during that period).

Issue #361, opened 2019-01-23
Switch JSON Handling to Jason


jherdman:
It seems like the broader Elixir community is settling on [Jason](https://hex.pm/packages/jason) for JSON encoding and decoding these days. Would it make sense to follow, or at least allow for a configurable encoder/decoder?

PR #362, opened 2019-01-25
Add json_library config with Jason default

chulkilee:
See #361


mix.exs
```diff
@@ -40,7 +40,8 @@ defmodule Exq.Mixfile do
     [
       {:elixir_uuid, ">= 1.2.0"},
       {:redix, ">= 0.8.1"},
-      {:poison, ">= 1.2.0 or ~> 2.0"},
+      {:poison, ">= 1.2.0 or ~> 2.0", optiona: true},
```
akira: Is this typo here, should be `optional`?

akira:
Thanks for PR!!  Looks good except one typo in mix file.  👍

mix.exs
akira: Will update this after merge as already looking at it.

akira:
Merging.   Thanks again!

abitdodgy:
Just curious, why does `poison` have to stay in the deps if handling is being switched to `Jason`? 

Wouldn't it make more sense to remove poison and leave it up to the user to add it to their own mix file if they want to use poison?

PR #366, opened 2019-02-24
Remove Inch from Dependencies

jherdman:
This appears to be completely unused, at least in CI. Perhaps the change should be, instead, to add it to the CI pipeline? If so I can close this PR and reopen one with that change.


akira:
👍 Sounds good, thanks!

Issue #199, opened 2016-07-08
Testing with Exq


hwhelchel:
- Fake Redis: Sidekiq has an [in memory fake queue](https://github.com/mperham/sidekiq/wiki/Testing#testing-worker-queueing-fake) that can be used for testing. It also allows for immediate running of jobs inline. Does exq have any similar testing functionality? Right now I can see the jobs going all the way through to redis and then I'm getting three errors in my test output. The jobs require reading data from the DB and the data isn't showing up. I think but am not sure this is because the foreground tests run in a sandboxed DB transaction.

I dug into this some more and found this [article](https://medium.com/@qertoip/making-sense-of-ecto-2-sql-sandbox-and-connection-ownership-modes-b45c5337c6b7#.fqfafo1bp) and [elixir issue](https://github.com/elixir-ecto/ecto/issues/1319)

I changed `Ecto.Adapters.SQL.Sandbox.mode(Arches.Repo, :manual)` to `Ecto.Adapters.SQL.Sandbox.mode(Arches.Repo, {:shared, self()})` which stopped error 1 and 3. But makes me confused because if `:shared` enables child processes to share parent connection and transaction then shouldn't Error 2 stop as well--the worker should be able to see the record in the DB?

Error 1:

```
10:55:44.325 [error] GenServer Exq terminating
** (stop) exited in: GenServer.call(Exq.Worker.Sup, {:start_child, ["{\"queue\":\"default\",\"jid\":\"ebae2772-86c9-4f94-b7af-f300c196df77\",\"enqueued_at\":1467989698.60931,\"class\":\"Arches.AnswerHandler\",\"args\":[\"18X86fmtfj9uvT7sWafKri2s3aa4RXxyQU\",2638,1567],\"retry\":true}", #PID<0.300.0>, "default", 872496, Exq.Stats, "exq", "Hoyts-MacBook-Air", Exq.Redis.Client, Exq.Middleware.Server]}, :infinity)
    ** (EXIT) no process
    (elixir) lib/gen_server.ex:596: GenServer.call/3
    (exq) lib/exq/manager/server.ex:269: Exq.Manager.Server.dispatch_job/3
    (exq) lib/exq/manager/server.ex:262: Exq.Manager.Server.dispatch_job/2
    (elixir) lib/enum.ex:1184: Enum."-map/2-lists^map/1-0-"/2
    (exq) lib/exq/manager/server.ex:229: anonymous fn/2 in Exq.Manager.Server.dequeue_and_dispatch/2
[...]

PR #374, opened 2019-04-14
Adding a queue adapter for testing purpose

samidarko:
## What changes were proposed in this pull request?

I'm facing for a while the same issue than Testing with Exq #199

I'm attempting to solve this issue by

- [x] Creating an adapter behavior **Exq.Adapters.Queue**
- [x] Creating a production queue **Exq.Adapters.Queue.GenServer**
- [x] Creating a testing queue **Exq.Adapters.Queue.Test**
- [x] Improve typespecs by making them more accurate with the expected parameters
- [x] Update documentation

The default queue is **Exq.Adapters.Queue.GenServer** and it's set in the config:

```Elixir
defmodule Exq.Support.Config do
  @default_config %{
    ...,
    queue_adapter: Exq.Adapters.Queue.GenServer
  }
```

If @akira is ok with this approach, I'll add some documentation and improve the typespecs in **Exq.Adapters.Queue** (I'll need a bit of guidance for the latter)

## How was this tested?

- [x] Unit tests
- [x] Manual tests

## How to test?

Replace `queue_adapter` with adapter of your choice

## Related Links
https://github.com/akira/exq/issues/199


akira:
Thanks for putting this together!  Looks like this could work for me,  @ananthakumaran any thoughts?

ananthakumaran:
The approach looks good overall.

lib/exq/adapters/queue/gen_server.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Exq.Adapters.Queue.GenServer do
```
ananthakumaran: we could probably use a better name here instead of genserver, which doesn't communicate anything and kind of misleading as this is not a gen server.

lib/exq/adapters/queue/test.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Exq.Adapters.Queue.Test do
+  @moduledoc false
+
+  @behaviour Exq.Adapters.Queue
+
+  def enqueue(_pid, _queue, worker, args, _options) do
+    {:ok, apply(worker, :perform, args)}
```
ananthakumaran: This is mostly a wish, not a requirement. We use `Exq.worker_job()` to get the job metadata like jid etc from within the worker. The test adapter would fail at that point. It would be great if we simulate that as well. We would have to do something similar to `Exq.Worker.Server#dispatch_work`.

ananthakumaran:
@samidarko It would also be nice if you could share what kind of patterns are used in your tests.

In our case, we enable Exq enqueuer in the test, but not the worker.

To test the enqueue side, We call assert on the length of queue.

```elixir
    assert Redix.command(Client, [
             "LLEN",
             "queue"
           ]) == {:ok, size}
```

To test the worker side, we call the perform directly and mock the worker_job.

```elixir
  setup_with_mocks([
    {Exq, [], [worker_job: fn -> %{jid: "123", queue: "high"} end]},
  ])
```

We mostly don't worry about the return value of the worker, just try to make assert on the things it's supposed to do.

I would be very happy to get rid of the mocks (we have to disable async because of mocks)

lib/exq/adapters/queue/gen_server.ex
samidarko: True. I could call it **Prod**, it would convey the environment as I called the other other one **Test**. Or maybe **Async** or  **Dispatch** ?

ananthakumaran: looking at the sidekiq api https://github.com/mperham/sidekiq/wiki/Testing, we could probably call it based on the functionality. Like redis, memory (will just keep the jobs in memory and we could later do assert/clear the queue), inline (will call the worker directly)

samidarko: I could rename **Exq.Adapters.Queue.Test** to **Exq.Adapters.Queue.Fake** and **Exq.Adapters.Queue.GenServer** to **Exq.Adapters.Queue.Async** ?

ananthakumaran: let's go with that

lib/exq/adapters/queue/test.ex
samidarko: If we want to avoid a `if Mix.env() == :test` in worker_job or another sort of "adapter". Maybe we could bring **Exq**._worker_job_ from **Exq.Enqueuer.EnqueueApi**. The `use` would contain:

```elixir
  def worker_job(name \\ nil, pid \\ self()) do
    queue_adapter = Config.get(:queue_adapter)
    queue_adapter.worker_job(name, pid)
  end
```
In **Fake** queue:
```elixir
  def worker_job(name \\ nil, pid \\ self()) do
    %{jid: "123", queue: name}
  end
```

In **Async** queue the normal implementation:
```elixir
  def worker_job(name \\ nil, pid \\ self()) do
    metadata = Metadata.server_name(name)
    Metadata.lookup(metadata, pid)
  end
```

I understand it breaks the semantic of the queue which is about queue operations but it seems to be a small code change. I'm open to other ideas. Maybe you could explain more what you had in mind when you said:
> We would have to do something similar to Exq.Worker.Server#dispatch_work.

ananthakumaran: I was thinking more on the lines of calling `Metadata.associate` from here.

``` elixir
job = construct_job(args)
:ok = Metadata.associate(metadata, self(), job)
result = apply(worker_worker, :perform, job.args)
```

But, I haven't looked through the details, like whether the metadata server will be up during test etc

samidarko: I had a look to this code but not sure to know where you're heading. Let me know you're idea and I'll be happy to help if I can

samidarko:
It's not so much about the tests but more about the issue I have when I run my test suite #199

But to give you some exemple, we are using Exq to send notifications or a newsletter. I call a phoenix end-point and I want to assert the newsletter status has been updated in the newsletter or the notification has been created.

akira:
Pinging @samidarko  @ananthakumaran  what changes required to merge this?

samidarko:
I don't have anything to add if you're ok on your side @akira @ananthakumaran

ananthakumaran:
I don't have many free cycles nowadays for further review. Please go ahead if the interface looks good. We could fix any remaining issues in the future.

LukevanTricht:
Is this PR waiting on anything? We could really use this functionality for our automated tests.

Issue #383, opened 2019-07-11
Performance issue when using realtime stats (with ExqUI)


neslinesli93:
Hi,
I've recently been hit by some performance issues while using this library. In particular, when visualizing failed tasks, my redis cluster's CPU usage has gone to 100% for a few seconds.

It seems like the reason lies in the way ExQ implements the API for fetching failed tasks, as it can be seen [here](https://github.com/akira/exq/blob/c1caff8052c5176b716715a8bf813d6e87d98f37/lib/exq/redis/job_stat.ex#L123).

As stated inside Redis docs, the `KEYS` command should be avoided when possible, and instead use `SCAN`, which is cursored and doesn't block the whole instance while it's executing.

Thanks for your time!

ananthakumaran:
Closing this as the related PR has been merged

PR #384, opened 2019-07-22
Replace KEYS with a cursored call to SCAN for realtime stats

neslinesli93:
As explained [here](https://github.com/akira/exq/issues/383), I've implemented a cursored call using `SCAN` instead of `KEYS`, to avoid freezing redis production instances that could possibly store million of keys related to exq jobs.

I still have to find a way to easily test this, since I seem to have problems running the tests using a dockerized version of redis, along with the default `test.exs` (even though I've set the `REDIS_PORT` variable appropriately)

### DIFF
--- a/lib/exq/redis/job_stat.ex
+++ b/lib/exq/redis/job_stat.ex
@@ -118,17 +118,34 @@ defmodule Exq.Redis.JobStat do
   end
 
   def realtime_stats(redis, namespace) do
-    {:ok, [failure_keys, success_keys]} =
-      Connection.qp(redis, [
-        ["KEYS", JobQueue.full_key(namespace, "stat:failed_rt:*")],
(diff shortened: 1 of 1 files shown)


akira:
Just to verify: 

`I still have to find a way to easily test this,`

You've tested this locally, but just the unit tests are failing?

neslinesli93:
Hi!

I can't find a way to reliably run tests, however they seem to be passing in the CI... I was referring to [this line](https://github.com/akira/exq/blob/471047d95293b0166d01a0a07e11c9ba90678418/test/test_helper.exs#L110), which manually launch redis instances using `:os.cmd`, and it's not working with my setup since I'm using a containerized redis instance.

I think it would be easier to submit PR if the README was a little bit clearer on how to test the library...

akira:
Merging this, thanks for PR @neslinesli93!  🎆

Issue #350, opened 2018-11-29
Excessive memory consumption on Exq.Stats


pacoguzman:
Sorry if this is not the place for this but I'm facing an excessive memory consumption on `Exq.Stats` process

```erlang
iex(my_app@10.73.215.121)3> :rpc.call(:"my_app@10.187.201.14", :sys, :get_state, [Exq.Stats])
%Exq.Stats.Server.State{queue: {[], []}, redis: Exq.Redis.Client}
iex(my_app@10.73.215.121)4> :rpc.call(:"my_app@10.187.201.14", :recon, :info, [Exq.Stats])   
[
  meta: [
    registered_name: Exq.Stats,
    dictionary: [
      "$initial_call": {Exq.Stats.Server, :init, 1},
      "$ancestors": [Exq.Sup,MyApp.Supervisor, #PID<35122.447.0>]
    ],
    group_leader: #PID<35122.446.0>,
    status: :waiting
  ],
  signals: [
    links: [#PID<35122.513.0>],
    monitors: [],
    monitored_by: [],
    trap_exit: true
  ],
  location: [
    initial_call: {:proc_lib, :init_p, 5},
    current_stacktrace: [
      {:gen_server, :loop, 7, [file: 'gen_server.erl', line: 381]},
      {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 247]}
    ]
  ],
  memory_used: [
    memory: 1246410040,
    message_queue_len: 0,
    heap_size: 18481566,
    total_heap_size: 155801133,
    garbage_collection: [
      max_heap_size: %{error_logger: true, kill: true, size: 0},
      min_bin_vheap_size: 46422,
      min_heap_size: 233,
      fullsweep_after: 65535,
      minor_gcs: 1
    ]
  ],
  work: [reductions: 1408825464]
]
```

As you can see it takes around 1.2GB just after booting the application, we removed the Stats middleware but the problem is still there. Could you help us to find the issue?

ananthakumaran:
how many jobs do you process per second? The stats server accumulates the events in a queue and sends the stats every second (stats_flush_interval). I would normally expect the message_queue to build up when the stats server can't keep up, but it looks empty `message_queue_len: 0`. Once the stats middleware is removed, none of this should matter, as nobody will send events to the stats server.

pacoguzman:
[...]

Issue #321, opened 2018-05-24
Processing jobs during deploy


tetiana12345678:
First of all, before I even ask for feedback, I want to say thank you for exq! It's a great library!

Now, I am looking for a feedback on how we solve situations we experience with exq:

1. Situation 1.
On start-up of the node, exq will move any entries in the "backup" queue to the main queue. The issue is when we deploy new version of the app, we have a new instance of the app running, but with the same node id. So exq removes job from the "backup" queue and it's being performed by new instance, while the older node is still running the job which completes successfully. So we have job executed twice.
Solution:
As suggested in the docs, we can implement a unique node identifier, so then on each deployment we will not touch jobs for previous deployment node.

2. That leads us to situation 2.
Node 1 (older node) got terminated without finishing all it's job 😲
Solution:
Implement backup queue cleaner for previous node. Backup cleaner is a `GenServer` process which after configurable period of time wakes up, find previous `node_id` and moves all job belonging to that `node_id` into a main queue.

Question is, are there a better way to make sure that we don't re-queuing in-progress jobs during deployment?

akira:
Typically I don't think people have multiple workers running on the same machine at the same time. However, in your situation, it seems that is the way deployment is done. In that case, the workaround that you mentioned seems like a reasonable way to ensure that jobs don't execute multiple times.

ananthakumaran:
fixed by #392

PR #392, opened 2019-10-24
Heartbeat implementation

ananthakumaran:
refer #261 #350 #321

## Node Recovery

A Node can be stopped unexpectedly while processing jobs due to various reasons like deployment, system crash, OOM, etc. This could leave the jobs in the in-progress state. Exq comes with two mechanisms to handle this situation.

### Same Node Recovery

Exq identifies each node using an identifier. By default machine's hostname is used as the identifier. When a node comes back online after a crash, it will first check if there are any in-progress jobs for its identifier. Note that it will only re-enqueue jobs with the same identifier. There are environments like Heroku or Kubernetes where the hostname would change on each deployment. In those cases, the default identifier can be overridden

### Heartbeat

Same node recovery is straightforward and works well if the number of worker nodes is fixed. There are use cases that need the worker nodes to be autoscaled based on the workload. In those situations, a node that goes down might not come back for a very long period.

Heartbeat mechanism helps in these cases. Each node registers a heartbeat at regular interval. If any node misses 5 consecutive heartbeats, it will be considered dead and all the in-progress jobs belong to that node will be re-enqueued.

This feature is disabled by default and can be enabled using the following config.

```elixir
config :exq,
    heartbeat_enable: true,
    heartbeat_interval: 60_000,
    missed_heartbeats_allowed: 5
```


argl:
Is there anything happening with this PR? It would be nice to get rid of the same-node-recovery semantics.

ananthakumaran:
The PR is ready and I plan to merge it to master soon (probably in two weeks). In the meantime would appreciate it if others could review or help with testing this branch.

PR #396, opened 2019-11-03
continuation of #374

ananthakumaran:
#374


ananthakumaran:
@samidarko @LukevanTricht would appreciate if you guys could review the api.

LukevanTricht:
I like it, clean, simple, does the job it needs to do!

lib/exq/mock.ex
```diff
@@ -0,0 +1,175 @@
...
+    }
+  end
+
+  defp to_job([_pid, queue, _time_or_offset, worker, args, _options]) do
+    %Job{
+      jid: UUID.uuid4(),
+      queue: queue,
+      class: worker,
+      args: args,
+      enqueued_at: DateTime.utc_now()
```
sergey-chechaev, line 160: Why did you decide not to use time_or_offset and set DateTime.utc_now() for Mock?

ananthakumaran: I can't think of any specific reason. What your use case? do you depend on the timestamp for enqueue_at or enqueue_in in your tests? I am open to fix this if there are use cases I didn't consider.

sergey-chechaev: Yes, we need to check is "enqueued_at" equal Date Time that we set

ananthakumaran: fixed by https://github.com/akira/exq/pull/449

PR #401, opened 2019-12-03
cast queue level concurrency

### LINKED ISSUE #386: handling queue concurrency of invalid type
Exq does not validate or handle queue level concurrency if the value is of type string (non-integer). If we set a string as a concurrency value, Exq behaves as if it's set to infinity (because of how comparison between int and string works in Elixir/Erlang)

Ideally it should fail or try to convert string to integer. Global concurrency seems to handle these cases. 

Example:

iex -S mix run --no-start
```elixir
defmodule TestWorker do
  def perform, do: :ok
end

Exq.start_link(queues: [{"test", "0"}])
{:ok, _} = Exq.enqueue(Exq, "test", TestWorker, [])

:timer.sleep(500)

{:ok, 1} = Exq.Api.stats(Exq.Api, :processed)
```

ananthakumaran:
We currently cast the global concurrency config. Make the logic uniform by casting queue level concurrency config as well. fixes #386

### DIFF
--- a/lib/exq/support/opts.ex
+++ b/lib/exq/support/opts.ex
@@ -124,27 +124,27 @@ defmodule Exq.Support.Opts do
   end
 
   defp get_config_concurrency() do
-    case Config.get(:concurrency) do
-      x when is_atom(x) ->
-        x
-
-      x when is_integer(x) ->
-        x
-
-      x when is_binary(x) ->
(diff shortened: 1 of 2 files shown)


akira:
Thanks for the PR @ananthakumaran and continued help! Merged.

Issue #388, opened 2019-09-24
Create API for parallelize job Enqueue


mrcdll:
We are using Exq for processing events coming from http request.

We had some issue with Exq.Enqueuer timing out when we reached request spikes over 10k.
```
Elixir.ErlangError · Erlang error: {:timeout, {GenServer, :call, [Exq, {:enqueue, "gmail_webhook", Pathfinder.Gmail.WebhookPostJob, [args]}}
```

The problem was resolved pooling Exq.Enqueuer processes with poolboy.

Would you be interested into wrapping the feature into a dedicated interface?

ananthakumaran:
Looking at the current code, we seem to unnecessarily serialize the enqueue step. Exq.Enqueuer.Server makes a request to redis and waits for reply **within** the gen_server loop. It only keeps the redis pid and namespace in the state. Ideally we should try to reduce the time spent inside the singleton gen_server loop. We could use a Agent which stores the pid/namespace. From the caller process do a lookup to get the pid/namespace and then make the redis request. This would remove the bottleneck at the Enqueuer level and likely Redix process might become the next bottleneck.

mrcdll:
Correct, and Redis handled the load pretty well.

Is it fine if I go ahead and open a PR?

ananthakumaran:
Yes, please go ahead

sb8244:
I've done some work sharding Redix in the past. The [recommended approach](https://hexdocs.pm/redix/real-world-usage.html) is a name based pool. This works well when you're the direct Redix consumer, because you can pick one of the Redix and you get "pooling" in a way (I call this sharding).

Would a similar approach to the Enqueuer API be worth it? In that path, the Enqueuer remains serialized, but there is a sharded pool of Enqueuers, each with their own Redix connection. It's extremely important to have a pool of Redix connections, so this would be advantageous. I'm not sure how I'd go about sharding Redix unless it was configurable through Exq, maybe with a behavior.

It would be great if Exq handled all of this for the consumer, sort of like Phoenix PubSub does with its sharded process approach.

ananthakumaran:
@sb8244 agree with sharding. Native sharding support would be nice. But we should first go for the low hanging fruit and remove the unnecessary serialize step we are introducing at Exq.

[...]

PR #390, opened 2019-10-03
Remove unnecessary serialization of enqueue calls

ananthakumaran:
## master

```
Name              ips        average  deviation         median         99th %
enqueue          0.85         1.17 s     ±1.05%         1.17 s         1.20 s
mix run --no-start bench.exs  9.19s user 1.40s system 103% cpu 10.246 total
```
## redix-parallel

```
Name              ips        average  deviation         median         99th %
enqueue         46.09       21.70 ms    ±27.34%       23.85 ms       29.91 ms
mix run --no-start bench.exs  20.80s user 7.24s system 340% cpu 8.244 total
```

@mrcdll I did an informal benchmark with artificial delay introduced by toxiproxy. This branch performs lot better than the current version. Again, I tried to keep the code change minimal inorder to not break any things. Could you test this branch and see if it solves your issue?

fixes #388

PR #402, opened 2019-12-06
Redix named-based pool

IRog:
This is a follow-up to https://github.com/akira/exq/issues/388#issuecomment-536859156

I'm using the approach recommended in https://hexdocs.pm/redix/real-world-usage.html#name-based-pool this should help with performance. I also have all tests passing locally.


ananthakumaran:
Could you share benchmark results before and after the change?

 Also, redis pool might not be the desired option every time. For example, we ping redis using Exq's connection (Exq.Redis.Client) to make sure the connection is working inside the health check. Introducing pool might make it complicated for those who don't require pool.

IRog:
I didn't do any direct benchmark but had a case where I was enqueuing 4K jobs and the latency for that dropped from 250ms to 200ms.

ananthakumaran:
4K jobs per second? 200ms seems a bit high unless the network latency is also high. I would be interested to know if there is any other bottleneck. From earlier comments https://github.com/akira/exq/pull/390#issuecomment-538200949, I was expecting a  lot more than that.

For the pool implementation, is it possible to make it configurable? By default use the same approach we are using now (no breaking change including the Redix process name) and if the pool config is enabled (also pool size should be configurable as well), then Exq should use a  pool.

PR #417, opened 2020-05-28
Allow job IDs to be pre-specified

bradediger:
Currently, exq generates UUID-based job IDs randomly. We have an application which computes stable job IDs based on some of the job arguments, so needs to preassign its job IDs. 

This change allows the `:jid` option to be passed to `enqueue` to override the job ID. Also adjusts `Exq.Mock` to pass through the job ID in the same fashion.

Thank you!


lib/exq/mock.ex
```diff
@@ -141,19 +141,19 @@ defmodule Exq.Mock do
     {:noreply, state}
   end
 
-  defp to_job([_pid, queue, worker, args, _options]) do
+  def to_job([_pid, queue, worker, args, options]) do
```
ananthakumaran, line 144: are we making these functions public just to aid testing? I would rather change the assertions in `FakeModeTest` and `InlineModeTest`.  More like a functional tests
```elixir
assert {:ok, jid} == Exq.Enqueuer.enqueue(ExqE.Enqueuer, "default", ExqTest.PerformWorker, [], jid: jid)
```

bradediger: Good call; I've made these methods private again and tested through the enqueue API in `FakeModeTest` and `InlineModeTest`. Thank you for the fast review!

bradediger: That [Travis failure](https://travis-ci.org/github/akira/exq/jobs/692231419) looks transient but I think you have to be an owner to kick it... @ananthakumaran are you able to re-run that job? Thank you!

PR #420, opened 2020-07-07
chore: Don't log disconnects

isaacsanders:
I don't know if there is more we need to do to handle disconnects, but this is a noisy error that has no action item, so I figured the right way to handle it is to pass it back to the caller, and not log it, as the caller should have reconnect logic.


ananthakumaran:
@isaacsanders just to be clear, you want to drop this error because it's already logged by Redix library? Because this is an error, it should be logged at least once. I am ok with dropping it from Exq if it's already logged by Redix. Could you share a sample log, that might make the issue more clear to me.

isaacsanders:
This error doesn't tell us anything. If you are supervising your Redix process and retrying errors higher up in the call chain, then you have handled this error. As is, it doesn't do anything but create a red herring for people to track down. Disconnects are erroneous if they happen with high frequency, which I am sure your supervision tree or retry logic handles.

isaacsanders:
Alternatively, this might be worth examining: https://github.com/whatyouhide/redix/blob/main/pages/Reconnections.md

ananthakumaran:
> This error doesn't tell us anything. If you are supervising your Redix process and retrying errors higher up in the call chain, then you have handled this error

I am not asking about the retry logic and so on. This PR disables an error log message for a specific error (disconnect), from a user perspective, I want to understand if the message is actually noisy or not. Posting a sample set of error logs here would help me understand the issue better.

isaacsanders:
> I am not asking about the retry logic and so on. This PR disables an error log message for a specific error (disconnect), from a user perspective, I want to understand if the message is actually noisy or not. Posting a sample set of error logs here would help me understand the issue better.

Redix already broadcasts when it disconnects, so logging those events is extraneous: https://github.com/whatyouhide/redix/blob/main/pages/Telemetry.md

benjijones:
> I am not asking about the retry logic and so on. This PR disables an error log message for a specific error (disconnect), from a user perspective, I want to understand if the message is actually noisy or not.

Hi, co-worker of Isaac's here. The log message shows like this:
```
{"@timestamp":"2020-07-17T02:30:00.005","level":"error","message":"%Redix.ConnectionError{reason: :disconnected}","metadata":{"application":"exq","domain":["elixir"],"file":"lib/exq/redis/connection.ex","function":"handle_response/2","gl":"<0.4548.0>","line":189,"mfa":"{Exq.Redis.Connection, :handle_response, 2}","module":"Elixir.Exq.Redis.Connection","pid":"<0.12671.0>","time":1594953000005206}}
```
We get this with regular frequency (a few times a day), as the redis connection is dropped (and re-connected automatically) due to normal network blips, etc. Since we just retry the request, it doesn't have any impact on our system's behavior, and we don't even care that it happened.

It's not that the message itself is specifically noisy- it's clear what's happening. It's that as a library, since you are still returning the error back to the user, there's no reason to additionally log it. If you were handling the error case somehow (like automatically retrying the request with a new connection) in a way that meant the client of the library would never see that error, then I would agree with surfacing that information via an error log- but since you're not, then it makes sense to just defer to whatever the client wants to do with that information (and just pass back the error value without logging it).

As a comparison note, it's consistent with how the Redix library operates- they don't log errors  in this case for the client- they just return the error value back. Wouldn't it make sense to adopt the same policy as another library in the stack, and let the client decide what to do with the information?

PR #431, opened 2020-10-24
relax poison dependency version

### LINKED ISSUE #430: Bump dependencies.
It seems that `exq` is using a fairly outdated version of `poison`. It would be really good if it could support more recent versions, because recent versions of dependencies like `poison` are used in many other packages, making it a hard choice to make to pick `exq`.

ananthakumaran:
fixes #430


williamweckl:
I think this has never been released. Can this be released?

PR #442, opened 2021-03-17
Misc doc changes

kianmeng:
Besides other changes, this commit ensures the generated HTML doc for
HexDocs.pm will become the main source doc for this Elixir library which
leverage on ExDoc features.

List of changes:
* Set and use latest ex_doc
* Set readme as main html page
* Add changelog to html doc
* Fix markdowns
* Update gitignore
* Update formatter config
* Add license section
* Refactor project config
* Badges and more badges!

Screenshot:
Screenshot-20210317223524-2626x1452

PR #443, opened 2021-03-21
Improve Github Actions

dolfinus:
Hello.

I have some improvement of the CI pipeline you have.

FIrstly, ubuntu-latest version was recently upgraded from 18.04 to 20.04. This causes an error with Elixir installation step:
https://github.com/akira/exq/pull/442/checks?check_run_id=2131663989
So I've changed ubuntu version back to 18.04 until the installer will be updated. 18.04 will be supported until April 2023.

Secondly, there is an annoying fact about Github Actions - sometimes it runs actions twice. For example:
1. You just pushed a commit to a branch - testing workflow will run only once
2. Someone forked your repo and creates a pull request - same thing
3. You've created a new branch (feature, bugfix or release branch), pushed a commit here and also created a pull request to the main branch - two copies of workflow will be run

Running a workflow twice does not make any sense, it just takes twice the time without any useful side effect.

So in this PR I've also added a separated job of checking if another copy of workflow already started: https://github.com/fkirc/skip-duplicate-actions

This job allows to:
1. Skip the second copy of workflow in the case 3 (branch in original repo + PR to main branch). Cases 1 and 2 will work just the same as now
2. Skip workflow if someone changed only files which are not connected to package code, like README or CHANGELOG
3. Cancel already started workflow in the same PR or branch if someone pushed new commits to it.

Thirdly, I've added caching step for deps and build temp files to speed up CI process.
Fourthly, I've added latest Elixir version 1.11 into the build matrix.

Example of job run:
https://github.com/dolfinus/exq/actions/runs/674056926

What do you think about that?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,31 +1,103 @@
-on: [pull_request, push]
+name: Test
+
+on:
+  push:
+  pull_request:
+    branches:
+      - main
+
 jobs:
-  test:
+  check_duplicate_runs:
+    name: Check for duplicate runs
+    continue-on-error: true
     runs-on: ubuntu-latest
-    env:
-      MIX_ENV: test
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    name: OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}}
+    outputs:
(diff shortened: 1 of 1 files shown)


ananthakumaran:
Thanks for the PR, let me merge and check it.

PR #444, opened 2021-04-28
handle timeouts on middleware pipeline

ananthakumaran:
Currently, if any of the middleware fails, (notably
Exq.Middleware.Job), it prevents the execution of remaining
middlewares. This is highly problematic as Exq.Middleware.Manager is
essential for Exq to function properly.

In this specific case, if the redis operation in Exq.Middleware.Job
timeouts (or fails), then Exq fails to decrement the busy worker count
and with enough number of such failures, the worker eventually will
stop picking up new jobs.

This commit tries to improve the conditions by adding a retry for
`remove_job_from_backup` operation. Retrying is ok in this case as the
operation is idempotent

### DIFF
--- a/lib/exq/redis/job_queue.ex
+++ b/lib/exq/redis/job_queue.ex
@@ -150,7 +150,9 @@ defmodule Exq.Redis.JobQueue do
   end
 
   def remove_job_from_backup(redis, namespace, node_id, queue, job_serialized) do
-    Connection.lrem!(redis, backup_queue_key(namespace, node_id, queue), job_serialized)
+    Connection.lrem!(redis, backup_queue_key(namespace, node_id, queue), job_serialized, 1,
+      retry_on_connection_error: 3
+    )
   end
 
   def scheduler_dequeue(redis, namespace) do
--- a/lib/exq/redis/job_queue.ex
+++ b/lib/exq/redis/job_queue.ex
@@ -250,12 +250,14 @@ defmodule Exq.Redis.JobQueue do
     offset = Config.backoff().offset(job)
     time = Time.offset_from_now(offset)
     Logger.info("Queueing job #{job.jid} to retry in #{offset} seconds")
-    enqueue_job_at(redis, namespace, Job.encode(job), job.jid, time, retry_queue_key(namespace))
+
+    {:ok, _jid} =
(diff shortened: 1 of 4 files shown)


lib/exq/redis/job_queue.ex
```diff
@@ -75,7 +75,9 @@ defmodule Exq.Redis.JobQueue do
     score = Time.time_to_score(time)
 
     try do
-      case Connection.zadd(redis, scheduled_queue, score, job_serialized) do
```
akira, line 78: Possible to add duplicate message here?

ananthakumaran:
```
127.0.0.1:6379> ZADD scheduled 1 abc
(integer) 1
127.0.0.1:6379> ZRANGE scheduled 0 100
1) "abc"
127.0.0.1:6379> ZADD scheduled 1 abc
(integer) 0
127.0.0.1:6379> ZRANGE scheduled 0 100
1) "abc"
```
> The same element only exists a single time, no repeated elements are permitted

Calling ZADD multiple times itself is not a problem. Technically it's possible though to end up having multiple jobs

1. call ZADD (timeouts, but the operation succeeded)
2. another worker move the job from scheduled queue to work queue
3. retrying ZADD added the same job to scheduled queue again

The major risk is the time gap between 1 & 2. If the job is scheduled to run in 60 seconds in the future, less likely to happen. But if the job is scheduled to run in 1 second, this could happen.

ananthakumaran: The ideal way to solve this problem is by not creating duplicate jobs under any conditions, this might require storing more information. My opinion is that the patch improves the current behavior at least. 

Currently, if there is a timeout
1) the job will get stuck in the in-process queue (on worker restart, it will get re-enqueued again)
2) if the operation succeeded on the Redis server, a duplicated job will get created when the worker gets restarted. One from scheduled-queue, one from the recovery of the in-process queue on worker restart

akira:
I see, sure that makes sense. We can also say jobs are at least once guarantee for this. In the future we can also try adding a "deduplication" for jobs, maybe support deduplication ID, seems common thing in some queueing systems (SQS: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html)

akira (approved):
LGTM thanks!   👍

PR #457, opened 2021-10-07
Add queuing_timeout and execution_timeout option for job

bluzky:
This PR implement 2 timeout options for exq.

- `queuing_timeout` (in second): TTL for job in queue, timed out job will be
discard using a middle ware
- `execution_timeout` (in second): if execution time is longer than
`execution_timeout` the worker is killed and not retry any more.

This could be used as an example for those who want to implement timeout.
This PR does not included test and execution


bluzky:
@ananthakumaran  wow thank you for your simple solution

PR #439, opened 2021-01-28
Update Manager and stats to correctly store information about proceses . (Backport from PR #289)

yknx4:
This changes the Job definition to be able to use Sidekiq Web UI again, I just backported the changes from the pr #289 without the unrelated stuff like Heartbeat

> This fixes problem with Sidekiq Web UI which has changed API from version 4.
> Now we correctly add data about manager and processes to redis

Issue #288, opened 2017-10-10
Sidekiq 5 compatibility with UI


ondrejbartas:
Hi,

I found out that Sidekiq 5 has different usage of redis than previous version.

Will you be interested in PR with changes for it?

ananthakumaran:
fixed in master

PR #458, opened 2021-10-10
Sidekiq 5 UI compatibility

ananthakumaran:
Adds support for sidekiq 5 UI. Threads (aka concurrency) is set to 0 since we allow custom dequeuers which might allow jobs based on different constrain. Busy count should be accurate. The `Quiet` button would cause the node to unsubscribe from all queues.

## Code 

1. This change introduces the concept of worker nodes. The list of active worker nodes is tracked in a set `exq:processes`. For each worker node, some metadata like subscribed queues, number of busy jobs, start time etc are tracked. A new GenServer called `Exq.Node.Server` is introduced to send heartbeat and the metadata info to the redis every 5 seconds. It also cleans up the dead worker nodes' info from redis.
2. The concept of per worker signal is introduced. Currently we support `TSTP` (quiet), which will ask the worker node to unsubscribe all active queue subscription.
3. Per worker metadata is stored under `exq:{node_id}` hash and kept updated by the GenServer. This info will expire in 60 seconds (which should happen when the worker dies)
4. Per worker active jobs are stored in `exq:{node_id}:workers` hash. Each individual job will get removed once it's done. The whole hash is deleted on node startup and on dead node cleanup

## Upgrade Instructions

There are 2 types of breaking changes introduced by this PR.

1. We used to store the set of actives jobs under `exq:processes`, but the list of active worker nodes node_id is being stored now. The new version gracefully handles unknown entries (it will be cleaned up eventually by prune_dead_nodes). So upgrade should not be an issue. But the old code won't handle node_id (non json), so it can't be rolled back without first deleting the `exq:processes` set manually from redis console.
2. Some of the data structures used also got changed which will break ExqUI. Specially `Exq.Api.processes/1` will return struct with different attributes. Those who use ExqUI should plan the upgrade of both Exq and ExqUI. The next release of ExqUI will be compatible with the new struct.

screenshot from single node
image

screenshot from multiple nodes (using custom identifier functionality)
image

fixes #439 #289 #288

### DIFF
--- a/lib/exq/node/server.ex
+++ b/lib/exq/node/server.ex
@@ -1,0 +1,94 @@
+defmodule Exq.Node.Server do
+  use GenServer
+  require Logger
+  alias Exq.Support.Config
+  alias Exq.Support.Time
+  alias Exq.Redis.JobStat
+
+  defmodule State do
+    defstruct [:info, :interval, :namespace, :redis, :node_id, :manager, :workers_sup]
+  end
+
+  def start_link(options) do
+    node_id = Keyword.get(options, :node_id, Config.node_identifier().node_id())
+
+    GenServer.start_link(
+      __MODULE__,
+      %State{
+        manager: Keyword.fetch!(options, :manager),
+        workers_sup: Keyword.fetch!(options, :workers_sup),
+        node_id: node_id,
+        info: node_info(node_id),
+        namespace: Keyword.fetch!(options, :namespace),
+        redis: Keyword.fetch!(options, :redis),
+        interval: 5000
+      },
+      []
+    )
+  end
+
+  def init(state) do
+    :ok = schedule_ping(state.interval)
+    {:ok, state}
+  end
+
+  def handle_info(
+        :ping,
+        %{
+          info: info,
+          namespace: namespace,
+          node_id: node_id,
+          redis: redis,
+          manager: manager,
+          workers_sup: workers_sup
+        } = state
+      ) do
+    {:ok, queues} = Exq.subscriptions(manager)
+    info = %{info | queues: queues}
+    busy = Exq.Worker.Supervisor.workers_count(workers_sup)
+
+    :ok =
+      JobStat.node_ping(redis, namespace, node_id, info, queues, busy)
+      |> process_signal(state)
+
+    :ok = schedule_ping(state.interval)
+    {:noreply, state}
+  end
+
+  def handle_info(msg, state) do
+    Logger.error("Received unexpected info message in #{__MODULE__} #{inspect(msg)}")
+    {:noreply, state}
+  end
+
+  defp process_signal(nil, _), do: :ok
+
+  defp process_signal("TSTP", state) do
+    Logger.info("Received TSTP, unsubscribing from all queues")
+    :ok = Exq.unsubscribe_all(state.manager)
+  end
+
+  defp process_signal(unknown, _) do
+    Logger.warn("Received unsupported signal #{unknown}")
+    :ok
+  end
+
+  defp schedule_ping(interval) do
+    _reference = Process.send_after(self(), :ping, interval)
+    :ok
+  end
+
+  defp node_info(node_id) do
+    {:ok, hostname} = :inet.gethostname()
+
+    %{
+      hostname: to_string(hostname),
+      started_at: Time.unix_seconds(),
+      pid: System.pid(),
+      tag: "",
+      concurrency: 0,
+      queues: [],
+      labels: [],
+      identity: node_id
+    }
+  end
+end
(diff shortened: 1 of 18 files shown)


lib/exq/node/server.ex
```diff
@@ -0,0 +1,103 @@
...
+      ) do
+    {:ok, queues} = Exq.subscriptions(manager)
+    busy = Exq.Worker.Supervisor.workers_count(workers_sup)
+    node = %{node | queues: queues, busy: busy, quiet: Enum.empty?(queues)}
+
+    :ok =
+      JobStat.node_ping(redis, namespace, node)
+      |> process_signal(state)
+
+    if Integer.mod(state.ping_count, 10) == 0 do
```
akira, line 63: What is 10 for?  Should be configured elsewhere or constant?

akira:
Possible to add more description of changes in PR / commit such as adding new GenServer for Node, changes to JobStat, etc.

akira:
Anything special instructions needed when upgrading to this branch once it lands? For handling existing inflight jobs / working nodes, job stats, etc..

lib/exq/node/server.ex
ananthakumaran: We have to remove the dead nodes from the processes set, the mod 10 is just a optimization to avoid doing the cleanup check on each ping instead do it on every 10th ping. The number 10 itself comes from [sidekiq implementation](
https://github.com/mperham/sidekiq/blob/46f4c6bd9ee0b2f5502a1dc0d03414989ad65063/lib/sidekiq/scheduled.rb#L160). If you feel this should be configurable, I can move it to application config

lib/exq/node/server.ex
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  def handle_info(msg, state) do
+    Logger.error("Received unexpected info message in #{__MODULE__} #{inspect(msg)}")
+    {:noreply, state}
+  end
+
+  defp process_signal(nil, _), do: :ok
+
+  defp process_signal("TSTP", state) do
```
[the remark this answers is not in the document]
ananthakumaran: The `Quiet` button in the UI pauses the worker (the worker stops picking up any new job). This is done via a signal list. UI will push `TSTP` string into the list and the worker will pop that and perform the action. I plan to add same feature in the Exq UI later as well.

ananthakumaran:
> Anything special instructions needed when upgrading to this branch once it lands? For handling existing inflight jobs / working nodes, job stats, etc..

If ExqUI is not used, then nothing needs to be done. There are some breaking changes related to where which info is stored, but the `prune_dead_nodes` parts will take care of the migration (it will remove the old job entries from `exq:processes` set). Though, if the upgrade can't be rolled back easily. As the old code will expect jobs in `exq:processes` set. To rollback the `exq:processes` needs to be deleted (we can include this instruction as part of release info)

If ExqUI is used, then the user has to wait for the new ExqUI release and upgrade ExqUI after Exq upgrade.

ananthakumaran:
> Possible to add more description of changes in PR / commit such as adding new GenServer for Node, changes to JobStat, etc.

I have updated the PR description with more info. Let me know if anything else needs to be covered

akira:
Thanks for the updates! Looks good I think - is the other PR also ready in ExqUI or still pending work on it?

ananthakumaran:
> Thanks for the updates! Looks good I think - is the other PR also ready in ExqUI or still pending work on it?

https://github.com/akira/exq_ui/pull/102 is ready for review. It includes the busy tab and related changes introduced in this branch.

akira:
@ananthakumaran Great thanks, will take a look at the other one as well, sorry for delay

akira:
> > Thanks for the updates! Looks good I think - is the other PR also ready in ExqUI or still pending work on it?
> 
> [akira/exq_ui#102](https://github.com/akira/exq_ui/pull/102) is ready for review. It includes the busy tab and related changes introduced in this branch.

Sounds good, I tried it out and works well!  🎉    Let me know your preference for release timing on these.  For the exq_ui maybe some additional unit tests (can be separate PR also).

ananthakumaran:
Both PRs can be merged to master. As for release, we could take a bit more time. I will open separate PRs for CI setup and tests for exq_ui. We will also try to get this tested out in our prod service (which is getting delayed due to phoenix version dependency)

akira:
> Both PRs can be merged to master. As for release, we could take a bit more time. I will open separate PRs for CI setup and tests for exq_ui. We will also try to get this tested out in our prod service (which is getting delayed due to phoenix version dependency)

Sounds good!  I added as maintainer to the other project as well, please feel free to merge.

akira:
This has been released, Exq 0.16.0 and ExqUI 0.1.0 🎉

PR #475, opened 2022-10-15
setup default redis timeout config

### LINKED ISSUE #474: expected :timeout to be an integer of :infinity, got: nil
Hi there,

I am in the process of trying `exq` and it looks very promising so far - however I stumbled upon an issue which prevents starting `exq` on application startup. 

I am getting the following Error:

```iex
** (Mix) Could not start application __REDACTED__: __REDACTED__.Application.start(:normal, []) returned an error: shutdown: failed to start child: Exq
    ** (EXIT) shutdown: failed to start child: Exq.Manager.Server
        ** (EXIT) an exception was raised:
            ** (RuntimeError) 


====================================================================================================
ERROR! Could not connect to Redis!

Configuration passed in: [[host: "localhost", port: __REDACTED__, database: 0, password: nil, name: Exq.Redis.Client, socket_opts: []]]
Error: :error
Reason: %ArgumentError{message: "expected :timeout to be an integer of :infinity, got: nil"}

Make sure Redis is running, and your configuration matches Redis settings.
====================================================================================================

                (exq 0.16.2) lib/exq/manager/server.ex:437: Exq.Manager.Server.check_redis_connection/1
                (exq 0.16.2) lib/exq/manager/server.ex:179: Exq.Manager.Server.init/1
                (stdlib 4.1.1) gen_server.erl:851: :gen_server.init_it/2
                (stdlib 4.1.1) gen_server.erl:814: :gen_server.init_it/6
                (stdlib 4.1.1) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
```

My Redis instance is up and running and reachable, the issue seems to be related to Exq's usage of Redix.
I already tried both starting `exq` as an OTP application as well as setting `start_on_application: false` and starting `exq` in my supervisor tree, the issue remains the same.

I also tried adding `timeout` to the `redis_options` parameter in  Exq's options.

Elixir Version: Erlang/OTP 25 Elixir 1.14.0 (compiled with Erlang/OTP 25)
Exq Version: 0.16.2
Redix Version: 1.2.0
OS: Debian 11 & macOS 13

Is anyone else having this issue?

ananthakumaran:
Earlier there was no default timeout value, we would send timeout: nil to Redix. With Redix 1.2, the option validation has become stricter and doesn't allow nil for timeout

fixes #474


fekle:
I had a look at the `redix` code, and it does seem a bit strange:

(https://github.com/whatyouhide/redix/blob/v1.2.0/lib/redix.ex)
```elixir
  @default_timeout 5000
[...]
     case Keyword.get(opts, :timeout, @default_timeout) do
        int when is_integer(int) ->
          int

        :infinity ->
          :infinity

        other ->
          raise ArgumentError,
                "expected :timeout to be an integer of :infinity, got: #{inspect(other)}"
      end
[...]
```

Looking at this code, passing `nil` as the timeout value _should_ work, given that `Keyword.get` gets a fallback value?
Maybe I'm missing something here?

ananthakumaran:
> given that Keyword.get gets a fallback value? Maybe I'm missing something here?

The fallback is used when the key is not present, not if the key is present and has nil as value

```elixir
iex(1)> Keyword.get([], :x, 10)
10
iex(2)> Keyword.get([x: nil], :x, 10)
nil
iex(3)>
```

fekle:
Ah whoops, I completely missed that. Thanks!

PR #481, opened 2022-12-02
Fix namespaced keys in scripts for ACL redis

korialis:
### Issue Details

It appears that when using the `EVAL` command redis checks all input keys against the ACL even if those keys are not used at all in the script. This makes the current version (`v0.17.0`) of `exq` raise a permission error in ACL protected redis clusters when enqueue is called since the library passes some non namespaced values as keys (including the possibly `nil` `unique_key`).

Below you can find some examples of this behaviour tested directly in an ACL protected `redis-cli`. 
The examples assume that a `namespace:` namespace is valid while `foo:` is not.

Passing a non namespaced, non used, key:

```
host:6379> EVAL "return ARGV[1]" 1 namespace: test
"test"
host:6379> EVAL "return ARGV[1]" 1 foo: test
(error) NOPERM this user has no permissions to access one of the keys used as arguments
```
Passing a nil value:

```
host:6379> EVAL "return ARGV[1]" 1 nil test
(error) NOPERM this user has no permissions to access one of the keys used as arguments
```
Passing a (not) namespaced but using a computed key.

```
host:6379> EVAL "return redis.call('SET', 'namespace:' .. 'queues', ARGV[1])" 1 foo test
(error) NOPERM this user has no permissions to access one of the keys used as arguments
host:6379> EVAL "return redis.call('SET', 'namespace:' .. 'queues', ARGV[1])" 1 namespace: test
OK
```

Citing this last example in relation to this disclaimer in the [docs](https://redis.io/docs/manual/programmability/eval-intro/):

> Important: to ensure the correct execution of scripts, both in standalone and clustered deployments, all names of keys that a script accesses must be explicitly provided as input key arguments. The script should only access keys whose names are given as input arguments. Scripts should never access keys with programmatically-generated names or based on the contents of data structures stored in the database.

### PR contribution
This pr aims to fix this refactoring how keys passed into the scripts, and avoiding passing `nil` keys to `Script.eval!/4`.

### DIFF
--- a/lib/exq/redis/job_queue.ex
+++ b/lib/exq/redis/job_queue.ex
@@ -35,12 +35,14 @@ defmodule Exq.Redis.JobQueue do
     try do
       [unlocks_in, unique_key] = unique_args(namespace, job, options)
 
+      keys = [full_key(namespace, "queues"), queue_key(namespace, queue)] ++ (unique_key || [])
+
       response =
         Script.eval!(
           redis,
           :enqueue,
-          [queue, full_key(namespace, ""), unique_key],
-          [job_serialized, job.jid, unlocks_in]
+          keys,
+          [queue, job_serialized, job.jid, unlocks_in]
         )
 
       case response do
@@ -96,8 +98,10 @@ defmodule Exq.Redis.JobQueue do
     try do
       [unlocks_in, unique_key] = unique_args(namespace, job, options)
 
+      keys = [scheduled_queue] ++ (unique_key || [])
+
       response =
-        Script.eval!(redis, :enqueue_at, [scheduled_queue, unique_key], [
+        Script.eval!(redis, :enqueue_at, keys, [
           job_serialized,
           score,
           jid,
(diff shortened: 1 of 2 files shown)


ananthakumaran:
Your patch looks good to me, I will merge it. I will check if all the other scripts are using the KEYS correctly later. 

> both in standalone and clustered deployments

One of the reasons I have ignored the note about the KEYS is because sidekiq (and in extension exq) is designed to run on a single redis server (or sentinel setup). I didn't realize KEYS would also affect ACL.

korialis:
> Your patch looks good to me, I will merge it. I will check if all the other scripts are using the KEYS correctly later.
> 
> > both in standalone and clustered deployments
> 
> One of the reasons I have ignored the note about the KEYS is because sidekiq (and in extension exq) is designed to run on a single redis server (or sentinel setup). I didn't realize KEYS would also affect ACL.

Thanks for merging the PR :) 

I did a quick check on the other scripts and it looked like they are correct. I had some doubts on  `scheduler_dequeue_jobs` but it looks ok since it passes a queue key and a namespace prefix, however a more expert eye would surely be helpful :)

Issue #482, opened 2023-01-20
Feature Request: enqueue a batch of jobs atomically


meysius:
# Usecase
In our application we need to send ~2000 events on our message broker all at once (or not at all, a.k.a atomically). So we need to submit many jobs all at once but it cannot be a single job so we can process them concurrently.

# Proposal
This can be achieved by using [Redis Transactions](https://redis.io/docs/manual/transactions/)

# Proposed API
```elixir
jobs = [
  [job_1_queue, job_1_worker, job_1_args],
  [job_2_queue, job_2_worker, job_2_args],
  [job_3_queue, job_3_worker, job_3_args],
  ...
]
{:ok, ack} = Exq.Enqueuer.enqueue_all(Exq.Enqueuer, jobs)
```

I will be happy to work on this myself and put a PR up, just wanted to make sure this feature is compatible with all the existing features or if there is anything I need to know before starting to work on this.

ananthakumaran:
Feature wise, this is a nice addition. I would suggest including options. It's ok to not support all the options supported by enqueue in the first cut, but we should include options, so support can be added in the future.

> [job_1_queue, job_1_worker, job_1_args, job_1_options],

> I will be happy to work on this myself and put a PR up

Thanks.

ericdude4:
@ananthakumaran can you release a new version to hex with this feature?

akira:
@ericdude4 just released a new version 👍

PR #483, opened 2023-01-21
Add enqueue_all for enqueuing a batch of jobs atomically

meysius:
This PR addresses: 
https://github.com/akira/exq/issues/482

API:
```elixir
jobs = [
  [queue_1, time_1, worker_1, args_1, options_1],
  [queue_2, time_2, worker_2, args_2, options_2],
  [queue_3, time_3, worker_3, args_3, options_3]
  ...
]

Exq.Enqueuer.enqueue_all(Exq.Enqueuer, jobs)

# Example returns
# 1.
{:error, reason}

# 2.
[
  {:ok, jid_1},
  {:ok, jid_2},
  {:ok, jid_3}
]

# 3.
[
  {:ok, jid_1},
  {:conflict, jid_2},
  {:conflict, jid_3}
]
```

@ananthakumaran Here is a quick draft, please take a quick look and let me know if I am missing anything obvious, I will get back to polishing, adding tests etc once I get an inital thumbs up.

cheers!

### DIFF
--- a/lib/exq/enqueue_api.ex
+++ b/lib/exq/enqueue_api.ex
@@ -98,6 +98,20 @@ defmodule Exq.Enqueuer.EnqueueApi do
         queue_adapter = Config.get(:queue_adapter)
         queue_adapter.enqueue_in(pid, queue, offset, worker, args, options)
       end
+
+      @doc """
+      Schedule multiple jobs to be atomically enqueued at specific times
+
+      Expected args:
+      * `pid` - PID for Exq Manager or Enqueuer to handle this
+      * `jobs` - List of jobs
+      #{@options_doc}
+
+      """
+      def enqueue_all(pid, jobs) do
+        queue_adapter = Config.get(:queue_adapter)
+        queue_adapter.enqueue_all(pid, jobs)
+      end
     end
   end
 end
--- a/lib/exq/enqueue_api.ex
+++ b/lib/exq/enqueue_api.ex
@@ -103,9 +103,23 @@ defmodule Exq.Enqueuer.EnqueueApi do
       Schedule multiple jobs to be atomically enqueued at specific times
 
       Expected args:
-      * `pid` - PID for Exq Manager or Enqueuer to handle this
-      * `jobs` - List of jobs
-      #{@options_doc}
+        * `pid` - PID for Exq Manager or Enqueuer to handle this
+        * `jobs` - List of jobs each defined as `[queue, worker, args, options]`
+          * `queue` - Name of queue to use
+          * `worker` - Worker module to target
+          * `args` - Array of args to send to worker
+          * `options`: Following job options are supported
+            * `max_retries` (integer) - max retry count
+            * `jid` (string) - user supplied jid value
+            * `unique_for` (integer) - lock expiration duration in seconds
+            * `unique_token` (string) - unique lock token. By default the token is computed based on the queue, class and args.
+            * `unique_until` (atom) - defaults to `:success`. Supported values are
+              * `:success` - unlock on job success
+              * `:start` - unlock on job first execution
+              * `:expiry` - unlock when the lock is expired. Depends on `unique_for` value.
+            * `schedule` - (optional) - used to schedule the job for future. If not present, job will be enqueued immediately by default.
+              * `{:in, seconds_from_now}`
+              * `{:at, datetime}`
 
       """
       def enqueue_all(pid, jobs) do
--- a/lib/exq/redis/job_queue.ex
+++ b/lib/exq/redis/job_queue.ex
@@ -121,23 +121,26 @@ defmodule Exq.Redis.JobQueue do
   end
 
   def enqueue_all(redis, namespace, jobs) do
-    { unique_keys, args } = extract_enqueue_all_keys_and_args(namespace, jobs)
+    {keys, args} = extract_enqueue_all_keys_and_args(namespace, jobs)
 
     Script.eval!(
       redis,
       :enqueue_all,
-      [ scheduled_queue_key(namespace) | unique_keys ],
+      [scheduled_queue_key(namespace), full_key(namespace, "queues")] ++ keys,
       args
     )
     |> case do
       {:ok, result} ->
-        result |> Enum.map(fn([status, jid]) ->
+        result
+        |> Enum.map(fn [status, jid] ->
           case status do
             0 -> {:ok, jid}
             1 -> {:conflict, jid}
           end
         end)
-      error -> error
+
+      error ->
+        error
     end
   end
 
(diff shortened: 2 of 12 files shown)


lib/exq/redis/job_queue.ex
```diff
@@ -120,6 +120,59 @@ defmodule Exq.Redis.JobQueue do
...
+  def enqueue_all(redis, namespace, jobs) do
+    schedule_queue = scheduled_queue_key(namespace)
+
+    job_attrs = extract_job_attrs(namespace, jobs)
+
+    lock_commands =
+      job_attrs
+      |> Enum.map(fn ({ jid, _score, _job_serialized, unique_key, unlocks_in }) ->
+        if unlocks_in do
+          ["SET", unique_key, jid, "PX", unlocks_in, "NX"]
```
ananthakumaran, line 132: The uniqueness feature can only be implemented via lua script, since we do a get, then lpush/zadd based on the get response.

I would start like this

```
MULTI // transaction
// load scripts to make sure  EVALSHA won't fail
SCRIPT LOAD // enqueue
SCRIPT LOAD // enqueue_at
EVALSHA hash key args
EVALSHA hash key args
EVALSHA hash key args
EVALSHA hash key args
EXEC
```

meysius: Thanks for quickly reviewing @ananthakumaran

From Redix Docs:
> ## Problems with transactions

 > There's an inherent problem with Redix's architecture and MULTI/EXEC
transaction. A Redix process is a single connection to Redis that can be used
by many clients. If a client A sends MULTI and client B sends a command before
client A sends EXEC, client B's command will be part of the transaction. This
is intended behaviour, but it might not be what you expect. This is why
transaction_pipeline/3 exists: this function wraps commands in MULTI/EXEC but
sends all in a pipeline. Since everything is sent in the pipeline, it's sent at
once on the connection and no commands can end up in the middle of the
transaction.

So I had to use Redix.transaction_pipeline/3 instead of MULTI/EXEC

> since we do a get, then lpush/zadd based on the get response.

I did the same exact thing,
for each job in the List I constructed a lock aquiring redis command:

```elixir
# If job option indicates uniqueness features is wanted, this command aquires a lock:
["SET", unique_key, jid, "PX", unlocks_in, "NX"]

# If uniqueness features are not wanted,
# this command returns an "OK" indicating the lock is not already aquired
["ECHO", "OK"]
```
so this construct a List of lock aquiring commands which I am passing to Redix.transaction_pipeline to run on redis.

from there, for each job we either have: `"OK"` or `nil` which then is used to construct another list of `ZADD` to push the job if lock_response is `"OK"` or `GET` to get the conflicted jid if the lock_response is `nil`

Finally that list of commands are also passed to Redix.transaction_pipeline and from there, for each job we either construct a `{ :ok, jid }` or a `{:conflict, conflict_jid}`

I quickly tested this locally, I confirmed uniqueness features are working fine, locally. but once I get a thumbs up, I ll start writing unit tests.

meysius: oh, wait, I just realized something. Are you saying this has to be a lua script because of the potential race condition between reading lock and enuquing the job?

ananthakumaran: > oh, wait, I just realized something. Are you saying this has to be a lua script because of the potential race condition between reading lock and enuquing the job?

Yes, btw, I don't fully understand how your approach works, but theoretically speaking it's not going to handle the race condition involved. Check the existing implementation, it uses  [lua script](https://github.com/akira/exq/blob/master/lib/exq/redis/job_queue.ex#L41) for enqueue

You would have to figure out how to do it inside multi/exec

ananthakumaran: > So I had to use Redix.transaction_pipeline/3 instead of MULTI/EXEC

transaction_pipeline is just `Redix.pipeline(conn, [["MULTI"]] ++ commands ++ [["EXEC"]], options)`

meysius: Yeah i am now reading in Redis website:
> Redis guarantees the script's atomic execution. While executing the script, all server activities are blocked during its entire runtime. These semantics mean that all of the script's effects either have yet to happen or had already happened.

https://redis.io/docs/manual/programmability/eval-intro/

I ll try writing it with Lua Script :)

lib/exq/redis/script.ex
```diff
@@ -51,6 +51,38 @@ defmodule Exq.Redis.Script do
         return {1, conflict_jid}
       end
       """),
+    enqueue_all:
```
meysius, line 54: Keys will come as:
```
[schedule_queue, job_1_unique_key, job_2_unique_key, ...]
```
And args will come as:
```elixir
[
  job_1_jid, job_1_score, job_1_serlialized_job, job_1_unlocks_in, # indexes 1, 2, 3, 4
  job_2_jid, job_2_score, job_2_serlialized_job, job_2_unlocks_in, # indexes 5, 6, 7, 8
 ...
]
```

meysius: According to https://redis.io/docs/manual/programmability/eval-intro/ I think we do not even need MULTI/EXEC as scripts are already run atomically.

meysius: @ananthakumaran please take a quick look when you can. Thanx :)

lib/exq/redis/job_queue.ex
```diff
@@ -685,15 +653,26 @@ defmodule Exq.Redis.JobQueue do
...
+      [unlocks_in, unique_key] = unique_args(namespace, job_data, unique_check: true)
+
+      # accumulating in reverse order for efficiency
+      {
+        [ unique_key | unique_keys_acc ],
+        [ unlocks_in, job_serialized, score, jid ] ++ job_attrs_acc
+      }
     end)
+
+    { Enum.reverse(unique_keys), Enum.reverse(job_attrs) }
```
meysius, line 676: reversing so the order maps the original order of jobs passed

lib/exq/redis/script.ex
```diff
@@ -51,6 +51,38 @@ defmodule Exq.Redis.Script do
...
+        local unlocks_in = tonumber(ARGV[(i - 1) * 4 + 4])
+        local unlocked   = true
+        local conflict_jid = nil
+
+        if unlocks_in then
+          unlocked = redis.call("set", unique_key, jid, "px", unlocks_in, "nx")
+        end
+
+        if unlocked then
+          redis.call('ZADD', schedule_queue, score, job)
```
ananthakumaran, line 74: There are two types of enqueue,
1) schedule in the future (enqueue_at)
2) enqueue now (enqueue)

ZADD is used for scheduling in the future, SADD/LPUSH is used for enqueue now

you might want to change it to SADD/LPUSH

ananthakumaran:
The overall direction looks good to me, you can get started with testing it out and other cleanups

lib/exq/redis/script.ex
meysius: The current version, expect jobs to be specified like `[queue, time, worker, args, options] = job` which by default expects time and supports scheduling for future. I was thinking if you pass `DateTime.utc_now` it will practically behave the same as enqueuing them right away. Although it may not be efficient to put jobs in schedule queue if they need to be enqueued right away.

I can make the `time` optional, and use SADD/LPUSH if time is not supplied. will see if I can also improve the API a bit.

ananthakumaran: There are two sets of apis: enqueue and enqueue_at/in. Though using `now` might seem the same, they are not implementation wise. Scheduled jobs sit in ZSET and are then picked up later by the scheduler. Look at the current behavior and use the same behavior for enqueue_all variant as well.

ananthakumaran:
```
jobs = [
  [queue_1, worker_1, args_1, options_1],
  [queue_2, worker_2, args_2, options_2],
  [queue_3, worker_3, args_3, options_3]
  ...
]
```

I think it might be better to remove the time and move it to options (in/at keys), since time is not required for normal enqueue (which would be the major use case)

meysius:
This PR is now ready for final review, Please let me know if anything is missing @ananthakumaran

lib/exq/enqueue_api.ex
```diff
@@ -98,6 +98,20 @@ defmodule Exq.Enqueuer.EnqueueApi do
         queue_adapter = Config.get(:queue_adapter)
         queue_adapter.enqueue_in(pid, queue, offset, worker, args, options)
       end
+
+      @doc """
+      Schedule multiple jobs to be atomically enqueued at specific times
+
+      Expected args:
+      * `pid` - PID for Exq Manager or Enqueuer to handle this
+      * `jobs` - List of jobs
```
ananthakumaran, line 107: could you include sample jobs spec (like the one in the README) here as well

ananthakumaran:
There is one job failure, I think you might have to run mix format. Apart from that, codewise, everything looks good to me. I will try to do some manual testing when I get time, hopefully by the end of this week, I would be able to merge this to master

lib/exq/enqueue_api.ex
meysius: sure

meysius: Screenshot 2023-01-23 at 11 43 44 PM

meysius:
I am not sure why my tests are not hiting `Exq.Adapters.Queue.Redis.enqueue_all` 
Coveralls report says enqueue_at is hit 4 times for example, but I was not able to find that function being called any different from enqueue_all in the exitting tests. So if you can let me know what I am doing wrong, I will write tests to hit those lines too.

meysius:
Also which format do you think makes more sense for the jobs: (jobs in `{}` vs in `[]`)
```
jobs = [
  {queue_1, time_1, worker_1, args_1, options_1},
  {queue_2, time_2, worker_2, args_2, options_2},
  {queue_3, time_3, worker_3, args_3, options_3}
  ...
]

jobs = [
  [queue_1, time_1, worker_1, args_1, options_1],
  [queue_2, time_2, worker_2, args_2, options_2],
  [queue_3, time_3, worker_3, args_3, options_3]
  ...
]
```

ananthakumaran:
> I am not sure why my tests are not hiting Exq.Adapters.Queue.Redis.enqueue_all

good catch, this is because we use Exq.Mock for tests. By default, it will just call the redis adapter, but for enqeueu_all we were translating to individual enqueue api calls. I have pushed a fix that should now fix the issue. The Redis adapter should have 100% coverage now.

> Also which format do you think makes more sense for the jobs: (jobs in {} vs in [])

I don't have any strong opinion on this, let's just go with the existing format.

ananthakumaran (approved):

meysius:
cool. lmk if anything else needed on my end. :) I just wait for this to be merged then.

PR #485, opened 2023-07-13
Bump version of Poison

frahugo:
Allow Poison 5.0.0 to be used.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - elixir: 1.6
-            otp: 20.3
           - elixir: 1.7
             otp: 21.3
           - elixir: 1.8.2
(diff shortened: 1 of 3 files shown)


ananthakumaran:
@frahugo do you know if I can find the changelog somewhere? I am happy to merge it, just wondering if there are any breaking changes that we need to know about.

frahugo:
> @frahugo do you know if I can find the changelog somewhere? I am happy to merge it, just wondering if there are any breaking changes that we need to know about.

@ananthakumaran The CHANGELOG is at the root. I ran the test locally with `Poison` as the Json library and tests were green. I guess it would not break anything as the json is used to serialize and deserialize data for the jobs. Maybe someone else who uses Poison can jump in?

ananthakumaran:
I was asking about poison changelog. Seems like they don't maintain any.

ananthakumaran:
@frahugo released

PR #493, opened 2024-08-22
Add job_id to Logger metadata for tracing

bluzky:
Hi,

I think it would be helpful to add job id to Logger metadata. So it would provide better log tracebility.

For example:
```elixir
config :logger, :console,
  format: "$time $metadata[$level] $message\n",
  metadata: [:job_id]
```
produce log
`13:13:00.120 job_id=8a762129-beff-4a2e-b9f6-35d1a1067581 [info] Time calculate pending orders - 0`

and we can find all log that belongs to the same job

How do you think.

### DIFF
--- a/lib/exq/worker/server.ex
+++ b/lib/exq/worker/server.ex
@@ -21,6 +21,7 @@ defmodule Exq.Worker.Server do
   alias Exq.Middleware.Server, as: Middleware
   alias Exq.Middleware.Pipeline
   alias Exq.Worker.Metadata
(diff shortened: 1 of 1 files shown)


ananthakumaran:
Couple of things to do
1) rename it to `jid`, which is what we use everywhere
2) do the same on `Exq.Middleware.Logger` as well in the `before_work`, so all the logs from exq logger will also have the correct jid and would be useful to filter out the logs of a single job with structured logging

Issue #494, opened 2024-12-09
Api.clear_* functions doesn't clear unique_tokens


dmitrykleymenov:
If jobs have `unique_token`s and we clear any queue, we have  `dead` tokens without a chance to remove them. The same is true for `remove_*` functions

ananthakumaran:
I believe the uniq tokens should have TTL set, so it should get cleared
eventually

dmitrykleymenov:
But what is the purpose of keeping tokens without jobs? For example, we have infinite TTL, cause we waiting for other services to up. And we can't remove tokens at all. May be an additional API function for removing it will do the trick. We just don't want to have "dead" values in Redis forever.

ananthakumaran:
> But what is the purpose of keeping tokens without jobs?

There might be some weird cases, but I generally agree with you. When a user deletes a job, they might expect the token to be deleted as well. Adding a separate API to clear tokens would be overkill; it can be an option in the same function and default to `true`.

Please feel free to open a PR.

ananthakumaran:
By the way, I am also not sure it's possible to delete **all** tokens since some tokens might be supplied by the user as well (not derived from job args). In that case, it's okay not to support those. Thinking again, it might be better to keep the default as `false` since there is no straightforward way to figure out if the job has unique tokens attached without doing additional redis operations.

PR #497, opened 2025-03-29
Add option to remove unique token

dmitrykleymenov:
According to [this issue](https://github.com/akira/exq/issues/494) Addition of `unlock` option to `remove_*_jobs` functions family

### DIFF
--- a/lib/exq/api.ex
+++ b/lib/exq/api.ex
@@ -291,13 +291,14 @@ defmodule Exq.Api do
   Expected args:
     * `pid` - Exq.Api process
     * `raw_job` - raw json encoded job value
+    * `unlock` - (boolean) whether to unlock jobs unique tokens, default is `false`
 
   Returns:
     * `:ok`
 
   """
-  def remove_failed_jobs(pid, raw_jobs) do
-    GenServer.call(pid, {:remove_failed_jobs, raw_jobs})
+  def remove_failed_jobs(pid, raw_jobs, unlock \\ false) do
+    GenServer.call(pid, {:remove_failed_jobs, raw_jobs, unlock})
   end
 
   def clear_failed(pid) do
--- a/lib/exq/api.ex
+++ b/lib/exq/api.ex
@@ -467,13 +467,14 @@ defmodule Exq.Api do
   Expected args:
     * `pid` - Exq.Api process
     * `raw_jobs` - list of raw json encoded job values
+    * `unlock` - (boolean) whether to unlock jobs unique tokens, default is `false`
 
   Returns:
     * `:ok`
 
   """
-  def remove_scheduled_jobs(pid, raw_jobs) do
-    GenServer.call(pid, {:remove_scheduled_jobs, raw_jobs})
+  def remove_scheduled_jobs(pid, raw_jobs, unlock \\ false) do
+    GenServer.call(pid, {:remove_scheduled_jobs, raw_jobs, unlock})
   end
 
   def clear_scheduled(pid) do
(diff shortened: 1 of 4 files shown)


lib/exq/api.ex
```diff
@@ -188,14 +188,15 @@ defmodule Exq.Api do
...
+    * `raw_jobs` - list of raw json encoded job values
+    * `unlock` - (boolean) whether to unlock jobs unique tokens, default is `false`
 
   Returns:
     * `:ok`
 
   """
-  def remove_enqueued_jobs(pid, queue, raw_jobs) do
-    GenServer.call(pid, {:remove_enqueued_jobs, queue, raw_jobs})
+  def remove_enqueued_jobs(pid, queue, raw_jobs, unlock \\ false) do
```
ananthakumaran, line 198: I think we should use `opts \\ []` instead of introducing the flag as a top-level argument. Could the name of the option be more descriptive? At first glance, `unlock` doesn't seem related to the uniqueness feature.

lib/exq/api/server.ex
```diff
@@ -155,8 +155,10 @@ defmodule Exq.Api.Server do
     {:reply, :ok, state}
   end
 
-  def handle_call({:remove_enqueued_jobs, queue, raw_jobs}, _from, state) do
+  def handle_call({:remove_enqueued_jobs, queue, raw_jobs, unlock}, _from, state) do
     JobQueue.remove_enqueued_jobs(state.redis, state.namespace, queue, raw_jobs)
+    if unlock, do: JobQueue.unlock_jobs(state.redis, state.namespace, raw_jobs)
```
ananthakumaran, line 160: Avoid this formatting style, it's ok to use 3 lines

lib/exq/api.ex
dmitrykleymenov: It's hard to select the right name. Could it be `unlock_uniqueness`? Or `clear_unique_tokens`?

ananthakumaran: `clear_unique_token` sounds good to me

lib/exq/redis/job_queue.ex
```diff
@@ -57,6 +57,10 @@ defmodule Exq.Redis.JobQueue do
     end
   end
 
+  def unlock_jobs(redis, namespace, raw_jobs) do
+    raw_jobs |> Enum.each(&unlock(redis, namespace, Job.decode(&1).unique_token))
```
ananthakumaran, line 61: Should we handle missing unique_token gracefully? Let's say we enable this flag by default on ExqUI, it's easier to just send a list of jobs and include this flag rather than trying to figure out which job has unique token etc. If there is no token, this could become a no-op?

ananthakumaran:
lgtm overall, I will test it a bit more and plan to merge this weekend

lib/exq/redis/job_queue.ex
dmitrykleymenov: Missed this comment somehow. In this case the token will be `nil` and we will end up with deleting the `"unique:"` key of the given namespace from the redis, which will silently say `"0"` and that's it. As far as I can see. And I think it's the most expected behavior. We can add a test for each queue with mixed jobs(with tokens and without).

ananthakumaran: I would prefer your second suggestion, which is more explicit about unique_token could be nil

dmitrykleymenov: What about tests? Should we add to `"remove enqueued jobs and clear their unique tokens"` , `"remove scheduled jobs and clear their unique tokens"` etc tests a job without the token to check it's removed as well?

dmitrykleymenov: Improved tests, added checks for removing jobs without tokens

PR #508, opened 2025-11-28
Fix Protocol.UndefinedError

neominik:
`Exq.Redis.JobStat.node_ping/3` returns nil when Redis returns an unexpected response, but `Exq.Node.Server.process_signals/2` doesn't handle nil and crashes when calling `Enum.each(nil, ...)`.


m-2k:
@ananthakumaran
@akira

Please add a tag that includes this merge request. We are getting errors on the production system related to this unreleased fix. Thanks!

ananthakumaran:
released

